PR #8, opened 2021-04-30
Publish on release actions

MRossol:
add conda packaging needs
add action files
add version file
update setup.py and README


dguittet (approved):
Thanks so much @MRossol ! I'll test this out with a dummy release.

Issue #12, opened 2021-09-15
.env file not setting API key


qualand:
Installation README instructions fail when setting NREL API key via .env file.  Myself and others had to set API key via set method.

barker59:
It appears this is happening in files where set_nrel_key_dot_env() has not been called prior to functions which request new resource data.

Suggested Fix:
call 'set_nrel_key_dot_env()' in the initialization of SiteInfo

PR #13, opened 2021-09-16
Issue #12 Fix: Updates README.md with better description of .env process.

barker59:
Also updates single_location.py by changing instances of "Lat" or "Lon" to "lat" and "lon", which improves consistency and may help prevent issues that can occur where the resource information is loaded for the default location, no matter where the user selects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,12 +37,19 @@ solar and storage.
     
     [https://developer.nrel.gov/signup/](https://developer.nrel.gov/signup/)
     
-7. Create the file `.env` in /HOPP and add one line to it, where \<key\> is the NREL API key you obtained:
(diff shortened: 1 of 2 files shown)


dguittet (changes requested):
Thanks for updating the README, but it seems like the issue Bill had was in reading the .env properly into HOPP despite having the correct file in place. That doesn't seem to be addressed here.

Also, your changes to single location have broken the tests-- take a look at the "Details":

image

It seems the error is from one of the run_hopp_calc tests.

barker59:
I definitely broke something in making those changes. single_location.py works as intended now despite the fact that the tests fail, whereas it didn't before, so will have to rewrite those tests also.

I'm going to scrap this pull request entirely and try again

PR #17, opened 2021-10-20
Update pysam

### LINKED ISSUE #16: Hybrid Depreciation and Taxes Not Copied over
Depreciation and tax variables are not copied over from technology component models into the hybrid financial models, leading to disconnects in those cash flow outputs

Need to update `calculate_financials` to handle these

dguittet:
Depreciation
- financial "hybridization" of Depreciation variables: using weight average for the allocation percents, and taking the union of each component's the deprecation types for the hybrid depreciation. This means, as an example, PV has 100MW allocated to straight line 5 year Fed bonus depreciation and Wind has 100MW allocated to straight line 15 yearFed bonus depreciation, then the Hybrid plant has 100MW allocated to each.

Debt, Construction, Reserve Accounts
- delete old 'construction_financing_cost_per_kw`
- initialize these variables to be 0 in powersource.py
- They should be set by the user
- fixes #16 

Update PySAM to version 3.0.0
 - add required battery financial model input: `batt_annual_discharge_energy`
 - turn off LCOS calculation
 - update battery replacement cost name to `om_batt_replacement_cost`
 - update which om types contribute the the om_expenses output
 - allows use of updated MHK models, @barker59


dguittet:
That's a good point about weighting by the total installed cost!

Also, I've been looking at the effect of adding the taxes to the dispatch which decreases the sales price. In the unit test, the battery performs better with taxes taken into account but in my larger example, it performs better when it pretends taxes are 0. I think this means I'm not understanding how the sales and purchases are being taxed by the financial model and so I'm tempted to remove the tax parameter...

dguittet:
I've updated the averaging and I also added in the option to charge the battery entirely from PV in order to qualify for the ITC credits.

I would appreciate if you could review those, thanks!

hybrid/hybrid_simulation.py
```diff
@@ -265,16 +265,26 @@ def calculate_financials(self):
...
-
         for v in generators:
             size_ratios.append(v.system_capacity_kw / hybrid_size_kw)
-        assert abs(sum(size_ratios) - 1) < 1e-7
 
-        def set_average_for_hybrid(var_name):
+        production_ratios = []
+        total_production = sum([v.annual_energy_kw for v in generators])
+        for v in generators:
+            production_ratios.append(v.annual_energy_kw / total_production)
```
qualand, line 274: How does the production_ratio work when the battery returns a negative value?

hybrid/hybrid_simulation.py
```diff
@@ -414,6 +414,15 @@ def simulate(self,
 
         logger.info(f"Hybrid Simulation complete. NPVs are {self.net_present_values}. AEPs are {self.annual_energies}.")
 
+    @property
+    def system_capacity_kw(self):
+        cap = self.outputs_factory.create()
+        for v in ("pv", "wind", "battery"):
```
qualand, line 420: Can we make this for loop based on power_sources so it will include future technologies?

dguittet: Good point! adding now

hybrid/hybrid_simulation.py
dguittet: I set up a `non_storage_production_ratio` for the O&M variable costs, which are computed differently for storage and non-storage generation. For PV and wind, the `om_production` is applied to the `gen` vector, and for battery, the `om_batt_variable_cost` is applied to just the battery discharge.

I added new tests to make sure these variable costs and the production tax credits make sense:
https://github.com/NREL/HOPP/blob/0d990492c12497f1a4b39b0ba8a5403be2805901/tests/hybrid/test_hybrid.py#L187

PR #23, opened 2022-03-09
Merge H2 Analysis

dguittet:


dguittet:
@barker59 

### These are some of the main conflicts in the `hybrid` folder that I worked through. There are more minor ones, please take in Changed Files

1. hybrid_simulation.py
    - rotor_diameter & hub_height inputs moved into wind_input dict / farm_config
    - `simulate_power` updated
        - power_flow_calculation: bool = False was removed until the power flow piece is merged in separately
    - `simulate_costs` function removed because it's duplicate of simulate
    - When passing storage sizes to financials, calculate the duration
   
2. wind_source.py
    - `total_installed_cost_dollars` function is in power_source.py
    - manual `system_capacity_kw` setter changes number of turbines

3. resource.py
    - Removed the API key rotation, this should be handled outside of resource.py which isn't responsible for API keys. The code using resource.py should wrap it in a try-catch or something

4. cost_calculator.py
    - change `storage_hours` from input to calculated from mwh and mw sizes

5. Consolidate `reopt_dev.py` and `reopt.py` with an `off_grid` parameter

### These are some things I looked through and decided not to merge for now:

1. `parameter_sweep.py` since the file it tries to load does not exist
2. `viz_data_v2` for same reason as 1
3. Everything power flow related since that needs its own set of testing and QA first
4. The standalone reopt files like `poller.py` since that's duplicate of REopt and I didn't see much use of it in the code
5. `jim_reopt_analysis.py` and `sample_reopt_development.py`
6. everything in `results/reopt_precompute`

#### Remaining Questions

1. How were all the files in the results folder generated? I think there should be some documentation of running what script with what inputs creates what output files where.

PR #26, opened 2022-04-14
Update API request protocol to V2

jmartin4nrel:
The developer network says that the original request protocols we have been using to download wind/solar resources are now deprecated, and to use the new V2 protocols.

Also: I have been unable to run anything from the current master branch without this change I made in elec_prices (essentially restoring a previous version). Someone with deeper knowledge should look into it


dguittet:
@jmartin4nrel Thanks for this update! Tested it out and the downloads work!

dguittet:
@jmartin4nrel Thanks for this update! Tested it out and the downloads work!

PR #32, opened 2022-05-11
Updates to integrate FLORIS v3

bayc:
This PR updates the custom wind module to integrate FLORIS v3. This vectorizes the wind series calculation and greatly speeds up the use of FLORIS for the wind plant modeling.

This implementation currently uses the farm defined in `floris_input.yaml`. A subsequent PR will include methods that allow the user to modify the wind plant being simulated.


dguittet:
Can you add floris to the requirements? the CI is failing since it cannot find floris

dguittet (approved):

PR #34, opened 2022-05-25
Create Constant Output Objective

dguittet:
Allow hybrid plant to be dispatched so that it produces the highest constant power level during the look-ahead period. This method does not use price or schedule, but only the power generation from pv, wind and/or csp.

The objective is selected using the `objective` key in the `dispatch_options` dictionary:

```
dispatch_options={ 
     'objective': 'constant_output',
     'n_look_ahead_periods': 48
     'n_roll_periods': 24
}
```
Here is an example output, from tests/hybrid/test_dispatch.py's `test_constant_output_dispatch`

image


dguittet:
The values of the "max constant output" per look-ahead period could be (but is not currently) sent to the Grid model as a curtailment limit, so any power produced above that is not sold to the grid, via a new `update_grid_curtailment` function in `hybrid_dispatch_manager.py` . Perhaps this function could be useful for other dispatch strategies as well. For instance, should we input the "desired_schedule" into PySAM's Grid's curtailment parameter? Or apply the "baseload schedule" as a interconnect limit so anything above it gets curtailed by the Grid model? 

The Grid model will automatically reduce the power output to the curtailed schedule while also doing the accounting of the total curtailed energy in a year. Thoughts @qualand ?  @genevievestarke ?

genevievestarke:
I think the grid curtailment adjustment would be really useful, but maybe as an option for the user rather than built in?  Or only in specific cases?  I'm not sure.  One of the control analysis options (the frequency regulation analysis) depends on the amount of power available above the baseload value but below the interconnection limit, so if we want to implement adjusted curtailment in the baseload case that analysis would just have to be adjusted a bit.

qualand (approved):

camirmas:
Closing for now, see #177

PR #35, opened 2022-05-26
Feature/desal model

barker59:
Merge new static pressure vessel model capabilities into feature/osw_h2 branch.

The main addition here is a static pressure vessel model which does not require CoolProp as a dependency.


barker59 (commented):
This looks great. Adds the needed static pressure vessel capabilities. Minimal dependencies affected and only one file change, so not much to check here before merging.

PR #36, opened 2022-05-31
New batt heuristic

genevievestarke:
Adding two new battery dispatch options (baseload one cycle battery heuristic and simple battery heuristic for power tracking) and a controls analysis after running the grid module.


parangat94:
@genevievestarke  and @dguittet  thanks for getting the tests to passing! Will review this next week.

hybrid/dispatch/hybrid_dispatch_builder_solver.py
```diff
@@ -501,14 +503,33 @@ def battery_heuristic(self):
...
             tot_gen = [wind + gen for wind, gen in zip(wind_gen, tot_gen)]
 
-        grid_limit = self.power_sources['grid'].dispatch.generation_transmission_limit
+        # grid_limit = self.power_sources['grid'].dispatch.transmission_limit
+        grid_limit = self.power_sources['grid'].dispatch.generation_transmission_limit        
+        if  'baseload' in self.options.battery_dispatch:
+            # Get difference between baseload demand and power generation and control scenario variables
+            if len(self.control) > 0:
+                control_case = list(self.control.keys())
+                baseload_limit = self.control[control_case[0]]['baseload_limit']
```
qualand, line 512: I think a class would clean this up

hybrid/dispatch/hybrid_dispatch_builder_solver.py
```diff
@@ -501,14 +503,33 @@ def battery_heuristic(self):
...
+                baseload_diff =  [(baseload_limit - x) for x in tot_gen]
+                self.power_sources['battery'].dispatch.baseload_difference = baseload_diff
+                goal_power = [baseload_limit/1000]*self.options.n_look_ahead_periods
```
qualand, line 515: `goal_power` is calculated up here but not used until later

hybrid/dispatch/hybrid_dispatch_builder_solver.py
```diff
@@ -501,14 +503,33 @@ def battery_heuristic(self):
...
-        if 'one_cycle' in self.options.battery_dispatch:
+
+            else:
+                raise ValueError("Baseload information must be input to run baseload battery case.")
+
+        elif 'one_cycle' in self.options.battery_dispatch:
             # Get prices for one cycle heuristic
             prices = self.power_sources['grid'].dispatch.electricity_sell_price
             self.power_sources['battery'].dispatch.prices = prices
+        #     # Get difference between baseload demand and power generation
```
qualand, line 526: I think this comment is misplaced

hybrid/dispatch/power_storage/one_cycle_battery_dispatch_heuristic_copy.py
```diff
@@ -0,0 +1,156 @@
...
+import pyomo.environ as pyomo
+from pyomo.environ import units as u
+
+import PySAM.BatteryStateful as BatteryModel
+import PySAM.Singleowner as Singleowner
+
+from hybrid.dispatch.power_storage.simple_battery_dispatch_heuristic import SimpleBatteryDispatchHeuristic
+
+
+class OneCycleBatteryDispatchHeuristicCopy(SimpleBatteryDispatchHeuristic):
```
qualand, line 11: I think you want to remove this file completely.

hybrid/dispatch/power_storage/one_cycle_baseload_battery_dispatch_heuristic.py
```diff
@@ -0,0 +1,158 @@
...
+import PySAM.BatteryStateful as BatteryModel
+import PySAM.Singleowner as Singleowner
+
+from hybrid.dispatch.power_storage.simple_battery_dispatch_heuristic import SimpleBatteryDispatchHeuristic
+
+
+class OneCycleBaseloadBatteryDispatchHeuristic(SimpleBatteryDispatchHeuristic):
+    """
+
+    fixed_dispatch: list of normalized values [-1, 1] (Charging (-), Discharging (+))
```
qualand, line 14: This should have a better description

qualand: I noticed that my original method doesn't have good description either

hybrid/dispatch/power_storage/one_cycle_baseload_battery_dispatch_heuristic.py
```diff
@@ -0,0 +1,158 @@
...
+import pyomo.environ as pyomo
+from pyomo.environ import units as u
+
+import PySAM.BatteryStateful as BatteryModel
+import PySAM.Singleowner as Singleowner
+
+from hybrid.dispatch.power_storage.simple_battery_dispatch_heuristic import SimpleBatteryDispatchHeuristic
+
+
+class OneCycleBaseloadBatteryDispatchHeuristic(SimpleBatteryDispatchHeuristic):
```
qualand, line 11: This class is mainly the `OneCycleBatteryDispatchHeuristic` we should use class inheritance to remove all of the repeated code.

hybrid/dispatch/power_storage/simple_battery_dispatch_heuristic_baseload.py
```diff
@@ -0,0 +1,149 @@
...
+                 pyomo_model: pyomo.ConcreteModel,
+                 index_set: pyomo.Set,
+                 system_model: BatteryModel.BatteryStateful,
+                 financial_model: Singleowner.Singleowner,
+                 fixed_dispatch: list = None,
+                 block_set_name: str = 'heuristic_baseload_battery',
+                 include_lifecycle_count: bool = False):
+        """
+
+        :param fixed_dispatch: list of normalized values [-1, 1] (Charging (-), Discharging (+))
```
qualand, line 26: We need to build out the doc strings for these heuristic classes

hybrid/dispatch/power_storage/simple_battery_dispatch_heuristic_baseload.py
```diff
@@ -0,0 +1,149 @@
+import pyomo.environ as pyomo
+from pyomo.environ import units as u
+
+import PySAM.BatteryStateful as BatteryModel
+import PySAM.Singleowner as Singleowner
+
+from hybrid.dispatch.power_storage.simple_battery_dispatch import SimpleBatteryDispatch
+
+
+class SimpleBatteryBaseloadDispatchHeuristic(SimpleBatteryDispatch):
```
qualand, line 10: This class should inherit the `SimpleBatteryDispatchHeuristic` class the `__init__` methods are exactly the same.

hybrid/hybrid_simulation.py
```diff
@@ -21,6 +21,7 @@
 from hybrid.layout.hybrid_layout import HybridLayout
 from hybrid.dispatch.hybrid_dispatch_builder_solver import HybridDispatchBuilderSolver
 from hybrid.log import hybrid_logger as logger
+from hybrid.controls import controls_analysis
```
qualand, line 24: I don't know if I am a fan of the way this is being integrated into `hybrid_simulation`...  Correct me if I am wrong but it seems like `controls_analysis` is just a way to pass information from outside of `hybrid_simulation` to the dispatch and then do some quick calculations and print some results. Passing information could be done through `dispatch_options`. The quick calculations and printing results could live in the `Grid` class. I am open to other ideas

tests/hybrid/test_dispatch.py
```diff
@@ -626,6 +626,26 @@ def test_hybrid_dispatch_heuristic(site):
     assert sum(hybrid_plant.battery.dispatch.charge_power) > 0.0
     assert sum(hybrid_plant.battery.dispatch.discharge_power) > 0.0
 
+def test_hybrid_dispatch_baseload_heuristic(site):
+    dispatch_options = {'battery_dispatch': 'baseload_heuristic',
+                        'grid_charging': False}
+    wind_solar_battery = {key: technologies[key] for key in ('pv', 'wind', 'battery')}
+    baseload_limit_kw = float(interconnect_mw * 0.75 * 1000)        # in kilowatts
+    baseload_percent = 90.0
+    simulation_options = {'control':{'baseload':{'baseload_limit':baseload_limit_kw, 'baseload_percent': baseload_percent}}}
```
qualand, line 635: In this test, we require the user to set both the `battery_dispatch` and the `control` dictionary. If we move these inputs into `dispatch_options` at least they will be defined right next to each other. We should also add a check to make sure the user has provide the correct keys if a specific `battery_dispatch` is selected.

tests/hybrid/test_dispatch.py
```diff
@@ -637,6 +657,23 @@ def test_hybrid_dispatch_one_cycle_heuristic(site):
     hybrid_plant.simulate(1)
 
     assert sum(hybrid_plant.battery.Outputs.P) < 0.0
+
+
+def test_hybrid_dispatch_one_cycle_heuristic_baseload(site):
```
qualand, line 662: Need to add test for the other control analysis method

qualand (changes requested):
@genevievestarke Great job! The majority of my comments were based on refactoring your additions to be more cleanly integrated into the existing code. I'm sorry if some of my comments come off harsh. I would be happy to discuss the requested changes. Please set something up my calendar is up-to-date.

dguittet:
I think @qualand hinted at this already, but I don't think hybrid/controls.py belongs in the hybrid folder, at least not in its current incarnation. That folder contains all the components to build up a hybrid simulation: the technology models, the financial models, the combination together, the control / dispatch classes that put that all together. `controls.py` if I'm understanding it correctly, looking like a kind of "metric" calculator, in that it's seeing how well the generation profile meets some ideal behavior.

One idea is to move the `baseload_sim` calculation into the associated dispatch class, [one_cycle_baseload_battery_dispatch_heuristic.py](https://github.com/NREL/HOPP/pull/36/files#diff-9065bf654f71ece7677079efe4cd0690bb1ef0ea4367d406b2121e8795223e37)

hybrid/hybrid_simulation.py
dguittet: I agree that the inputs to the new dispatch should be part of dispatch options, and that hybrid.controls should not be part of hybrid_simulation as they are more output metrics / analysis than fundamental to performing the hybrid simulation.

hybrid/dispatch/hybrid_dispatch_builder_solver.py
```diff
@@ -501,14 +507,31 @@ def battery_heuristic(self):
...
+            self.power_sources['battery'].dispatch.baseload_difference = baseload_diff
+            # Note: grid_limit and goal_power are in MW
+
+        elif 'one_cycle' in self.options.battery_dispatch:
             # Get prices for one cycle heuristic
             prices = self.power_sources['grid'].dispatch.electricity_sell_price
             self.power_sources['battery'].dispatch.prices = prices
 
-        self.power_sources['battery'].dispatch.set_fixed_dispatch(tot_gen, grid_limit)
+        if 'baseload' in self.options.battery_dispatch and 'one_cycle' not in self.options.battery_dispatch:
```
dguittet, line 529: can you make this more explicit? like if self.options.battery_dispatch == x, then do this, elif it equals y then do that. This is a bit confusing

hybrid/dispatch/hybrid_dispatch_builder_solver.py
```diff
@@ -501,14 +507,31 @@ def battery_heuristic(self):
...
+            if any(key not in self.options.baseload.keys() for key in required_keys):
+                is_missing = [key not in self.options.baseload.keys() for key in required_keys]
+                missing_keys = [missed_key for (missed_key, missing) in zip(required_keys, is_missing) if missing]
+                raise ValueError(type(self).__name__ + " requires the following keys: " + str(missing_keys))
+                
+            # Get difference between baseload demand and power generation and control scenario variables
+            baseload_value = self.options.baseload['limit']
+            baseload_diff =  [(baseload_value - x) for x in tot_gen]
+            self.power_sources['battery'].dispatch.baseload_difference = baseload_diff
+            # Note: grid_limit and goal_power are in MW
```
dguittet, line 522: are the units for the baseload dispatch options in MW or KW? in the Grid doc string, it says KW for `'baseload_limit': constant value of firm power you want to provide (kW)`?

dguittet:
I agree with @qualand on modifying the inheritance of the `OneCycleBaseloadBatteryDispatchHeuristic` and `BatteryBaseloadDispatchHeuristic` and removing all the duplicated code. I can go ahead and make those changes, while making sure the tests are still passing with that change.

Though it made me wonder about a more basic question, how does `one_cycle_baseload_heuristic` perform at meeting the baseload requirement? at first guess, I would think it wouldn't do very well if it can only cycle once a day. And in the case that it doesn't work well, should it even be merged?

tests/hybrid/test_dispatch.py
dguittet: I think both are tested, one in each of these two tests

hybrid/dispatch/hybrid_dispatch_builder_solver.py
genevievestarke: I changed this so it's hopefully a bit more clear

hybrid/dispatch/hybrid_dispatch_builder_solver.py
genevievestarke: The user input for firm power is in kW, but the inputs to the dispatch option 'baseload_heuristic' are in MW, which is what this note is trying to say (the code it describes got moved).  I moved the comment down to where the 'baseload_heuristic' options is actually called to clarify this point.

PR #47, opened 2022-08-05
created run file for degradation and failure

kbrunik:
Created failure class (failure.py)
Adds failure to generation from wind and pv.
Adds failure to battery used.
TODO: Add failure to hydrogen production.

Created run_degradation_failure.py
Combines degradation and failure class into one run file.
TODO: Add hydrogen production.


kbrunik:
Created a baseline scenario for HOMP land-based degradation and failure.
Outputs LCOE and LCOH with and without degradation and failure of each technology accounted for.

PR #69, opened 2022-11-12
SiteInfo loaded from KML file

dguittet:
- Reading a site from a KML file with exclusions
- Appending wind turbines and solar regions as placemarks to the KML file
- Write the updated KML file
- Tests in `tests/hybrid/test_layout.py` with `example_layout.kml`

@celine-carbonez @nimalgui Just FYI, no actions needed on your part


tests/hybrid/test_layout.py
```diff
@@ -145,4 +145,33 @@ def test_hybrid_layout_solar_only(site):
...
     for i in range(4):
         assert solar_region[i] == pytest.approx(expected_solar_region[i], 1e-3)
-        assert buffer_region[i] == pytest.approx(expected_buffer_region[i], 1e-3)
\ No newline at end of file
+        assert buffer_region[i] == pytest.approx(expected_buffer_region[i], 1e-3)
+
+
+def test_kml_file_read():
+    filepath = Path(__file__).absolute().parent / "layout_example.kml"
+    SiteInfo.kml_read(filepath)
```
Matthew-Boyd, line 153: Is this just testing that it doesn't raise an exception?

tests/hybrid/test_layout.py
```diff
@@ -145,4 +145,33 @@ def test_hybrid_layout_solar_only(site):
...
+    site = SiteInfo(site_data, solar_resource_file=solar_resource_file, wind_resource_file=wind_resource_file)
+
+    x = site.polygon.centroid.x
+    y = site.polygon.centroid.y
+    turb_coords = [x - 500, y - 500]
+    solar_region = Polygon(((x, y), (x, y + 5000), (x + 5000, y), (x + 5000, y + 5000)))
+
+    filepath_new = Path(__file__).absolute().parent / "layout_example2.kml"
+    site.kml_write(filepath_new, turb_coords, solar_region)
+    assert filepath_new.exists()
```
Matthew-Boyd, line 176: Maybe test that the file was (properly) written to as well?

Matthew-Boyd (approved):

tests/hybrid/test_layout.py
dguittet: The actual test is in line 159, but yes this line 153 will fail if an exception is raised. Is that what you mean?

dguittet: Oh nevermind, I see what you meant

PR #71, opened 2022-11-20
Feature/pressure vessel

jaredthomas68:
This pull request contains 

1. Minor changes to make it possible to run h2 analysis from an external directory

2. An adaptation of the pressure vessel model from Peng Peng at LBL as of 10/21/2022. It performs a curve fit against several model runs and then uses the result to approximate capex, opex, and energy. In the future it should be adjusted to not use an excel file for input and to get model results without the curve fit if desired. There are basic tests to make sure the results are reasonable as compared to the version received from Peng Peng. Sources are given in the specific readme in the directory with the pressure vessel model.

3. The pipe storage model has been moved to a common directory with the new pressure vessel model (h2_storage). In the future this location should be relocated, but I'm not sure where yet.

4. Some edits and bug fixes for the pressure vessel model based on discussions between the original author of the code, Jared Thomas, and Abhineet Gupta.


jaredthomas68:
@elenya-grant

PR #83, opened 2023-01-14
Adding detailed PV model (pvsamv1)

Matthew-Boyd:
This adds pvsamv1 functionality to HOPP and the ability to use a custom PV plant and/or layout model. Tests that demonstrate these capabilities are test_load_data() and test_hybrid_pv_plants() in test_detailed_pv.py

A simple custom financial model has not been fully implemented, but that is the next task.

This code was originally built off the pvplant branch in dguittet/HOPP, which has a different history than the detailed_pv_opt branch. Subsequently, the merge of detailed_pv_opt into pvplant_imp required additional fixes to the detailed model code. The detailed_pv_opt branch had multiple failing tests, which were not fixed, but all tests related to the detailed models now pass.

If the prior pvplant branch should not have been merged with detailed_pv_opt, than this pvplant_imp branch should not either and a rebase is instead necessary.


dguittet:
@Matthew-Boyd there are still 2 tests that aren't passing, due to "gen" not being assigned

dguittet:
@Matthew-Boyd I've just pushed an tiny edit to fix the `gen` issue. The problem was that the `value` function was trying to access that property of the system model before it had even been assigned, so I moved that line into the try catch block.

Now the issue I'm seeing is that `subarray1_rear_soiling_loss` is not found in the model. `subarray1_rear_soiling_loss` is not in PySAM 3.0.0, which is the official version HOPP is using. Can you please check what version of PySAM you're on?

Matthew-Boyd:
> @Matthew-Boyd I've just pushed an tiny edit to fix the `gen` issue. The problem was that the `value` function was trying to access that property of the system model before it had even been assigned, so I moved that line into the try catch block.
> 
> Now the issue I'm seeing is that `subarray1_rear_soiling_loss` is not found in the model. `subarray1_rear_soiling_loss` is not in PySAM 3.0.0, which is the official version HOPP is using. Can you please check what version of PySAM you're on?

@dguittet Thank you for catching this, I wasn't paying attention to the version. This latest commit should fix the tests.

tests/hybrid/test_detailed_pv.py
```diff
@@ -97,23 +146,98 @@ def test_load_data(site):
...
+        'fin_config': default_fin_config,
+        'pan_file': "",
+        'ond_file': "",
+    }
+
+    annual_energy_expected = 108239401.3
+
+    # Test standalone DetailedPVPlant 
+    pv_plant = DetailedPVPlant(site=site, pv_config=pv_config)
+    pv_plant.simulate_power(1, False)
```
dguittet, line 210: Layout fxn is needed here too. Normally the layout fxn is called from within HybridSimulation, but here we have to do it manually

hybrid/detailed_pv_plant.py
```diff
@@ -124,6 +140,7 @@ def system_capacity_kw(self, size_kw: float):
         :param size_kw:
```
dguittet, line 140: can you please add to the comment whether this is DC or AC?

hybrid/financial/custom_financial_model.py
```diff
@@ -76,6 +76,65 @@ class SystemCosts(FinancialData):
...
+    depr_itc_fed_sl_20: float
+    depr_itc_sta_sl_20: float
+    depr_bonus_fed_sl_39: float
+    depr_bonus_sta_sl_39: float
+    depr_itc_fed_sl_39: float
+    depr_itc_sta_sl_39: float
+    depr_bonus_fed_custom: float
+    depr_bonus_sta_custom: float
+    depr_itc_fed_custom: float
+    depr_itc_sta_custom: float
```
dguittet, line 121: while this is following the structure correctly, I don't think you need to include Incentives in your simple custom financial model, unless of course you know how to calculate them already and it's not hard / time-consuming.

hybrid/financial/custom_financial_model.py
```diff
@@ -76,6 +76,65 @@ class SystemCosts(FinancialData):
...
+    These financial inputs are used in simulate_financials in various files.
+    The names correspond to PySAM.Singleowner variables.
+    To add any additional system cost, first see if the variable exists in Singleowner, and re-use name.
+    This will simplify interoperability
+    """
+    ptc_fed_amount: float
+    ptc_fed_escal: float
+    itc_fed_amount: float
+    itc_fed_percent: float
+
```
dguittet, line 136: same with tax incentives

dguittet (commented):
I think it looks good so far! Just a couple minor changes

tests/hybrid/test_detailed_pv.py
```diff
@@ -89,6 +135,9 @@ def test_load_data(site):
```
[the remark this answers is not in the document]
Matthew-Boyd: You're right. I added a call to processed_assign() which is called by the detailed plant model's constructor.

tests/hybrid/test_detailed_pv.py
Matthew-Boyd: Fixed by the aforementioned call in processed_assign().

hybrid/detailed_pv_plant.py
Matthew-Boyd: Done (DC).

hybrid/financial/custom_financial_model.py
Matthew-Boyd: These are added because the Grid object currently uses the Singleowner model and there would otherwise be an error when it tries to aggregate the PV financial model parameters. I'll remove these later once I generalize the Grid financial model.

hybrid/financial/custom_financial_model.py
Matthew-Boyd: same reason as depreciation addition

hybrid/detailed_pv_plant.py
```diff
@@ -67,6 +67,7 @@ def processed_assign(self, params):
         self.assign(params)
         self.value('system_capacity',
                    get_num_modules(self._system_model) * get_module_power(self._system_model) * 1e-3)
+        self._layout.compute_pv_layout(self.system_capacity)
```
dguittet, line 70: Why did you add this compute call to processed_assign? Logically they seem like different tasks. Normally the layout is called by `HybridLayout` but I think there's quite a bit of mess and inconsistency in that code. I'd like to straighten that up in the near future so that it's easier / more intuitive to get to the layout computation, but for now, could we just add ` pv_plant._layout.compute_pv_layout(self.system_capacity)` to the test itself?

dguittet (approved):
I think this will be good to merge after moving the compute pv layout to the test itself

hybrid/financial/custom_financial_model.py
dguittet: Gotcha thanks

hybrid/detailed_pv_plant.py
Matthew-Boyd: I think I was enforcing coherence between the parameters. Maybe instead of calling compute_pv_layout(),  _compute_string_config() and _set_system_layout() should be called separately instead?

Matthew-Boyd:
> I think this will be good to merge after moving the compute pv layout to the test itself

I don't think this will be merged now that we're breaking up the PRs for the sake of their CI. I'm working now on adding this code to a separate branch off master.

hybrid/detailed_pv_plant.py
dguittet: Hmm... `_compute_string_config` and `_set_system_layout` both require design variable inputs though, whereas the params are fixed throughout

PR #88, opened 2023-01-31
Feature/offshore h2

nRiccobo:
Proposed updates to offshore-h2 branch by providing cost and sizing for a fixed-based platform for hydrogen production. File: fixed_platform_h2.py. It can run as a standalone and uses the example_fixed_platform_h2.yaml file as a configuration file. User must have ORBIT installed properly. Also included is a test_offshore.py test script run with pytest. 

You can use this code to append the design and install phases of an ORBIT project. Then you can aggregate results into system/installation capex. 

Alternatively, you could import this file into any script and use the provided functions to calculate the cost and size of a platform. Cost functions are informed by default values from ORBIT.


nRiccobo:
Updated at 2:50 MST on 1/31. Cleaned up code and removed test files per Jared and my discussion.

PR #99, opened 2023-02-09
Add detailed PV model (pvsamv1)

Matthew-Boyd:
- Improve value assignment in power_source
- Add detailed PV model (pvsamv1)
- Add custom PV model capability


dguittet:
@Matthew-Boyd I think the workflow goes: target design capacity and ratios -> figure out the nstrings, nmodules etc which are integer and so might not result in exactly the desired target design capacity and ratios -> then set the `system_capacity`, which is really more like a reporting step since in Pvsamv1, the `system_capacity` is a calculated variable. Did I understand the question correctly?

Matthew-Boyd:
> @Matthew-Boyd I think the workflow goes: target design capacity and ratios -> figure out the nstrings, nmodules etc which are integer and so might not result in exactly the desired target design capacity and ratios -> then set the `system_capacity`, which is really more like a reporting step since in Pvsamv1, the `system_capacity` is a calculated variable. Did I understand the question correctly?

@dguittet Thanks, and yes, you understood correctly. I wasn't sure if this would apply to plant initialization, but I guess so. I've made the change.

hybrid/detailed_pv_plant.py
```diff
@@ -0,0 +1,103 @@
...
+        n_strings, system_capacity, n_inverters = align_from_capacity(
+            system_capacity_target=self.value('system_capacity'),
+            modules_per_string=self.value('subarray1_modules_per_string'),
+            module_power=get_module_power(self._system_model) * 1e-3,
+            inverter_power=get_inverter_attribs(self._system_model)['P_ac'] * 1e-3,
+            n_inverters_orig=self.value('inverter_count')
+        )
+        self._system_model.SystemDesign.subarray1_nstrings = n_strings
+        self._system_model.SystemDesign.system_capacity = system_capacity
+        self._system_model.SystemDesign.inverter_count = n_inverters
```
dguittet, line 65: How I've tried to handle this issue of interdependent variables is shown in the wind source. For instance, in the `__init__`, the turbine rating is assigned from the config, and the rating then changes the system capacity:

https://github.com/NREL/HOPP/blob/c4e2a55976d18f3fd8f222604979035181448174/hybrid/wind_source.py#L71

The `turb_rating` is actually a Python class property with custom-defined getters and setters, as you can see here:

https://github.com/NREL/HOPP/blob/c4e2a55976d18f3fd8f222604979035181448174/hybrid/wind_source.py#L130

Whenever the setter `self.turb_rating = x` is called, we end up going to that custom function. That custom function's job is to make everything in PySAM consistent to the best of its ability.

When you have these new Pvsamv1 variables that users are supposed to use, you will need to add these class properties to control how they are accessed and modified. We want to enable the ability for the user to do:

```
pvplant.n_strings = 11
pvplant.module_power = 240
pvplant.subarray1_modules_per_string = 10
...
```

and have the end resulting Pvsamv1 model consistent.

hybrid/hybrid_simulation.py
```diff
@@ -148,7 +149,12 @@ def __init__(self,
             power_sources[k.lower()] = power_sources.pop(k)
 
         if 'pv' in power_sources.keys():
-            self.pv = PVPlant(self.site, power_sources['pv'])
+            if 'pv_plant' in power_sources['pv']:
+                self.pv = power_sources['pv']['pv_plant']                       # User instantiated plant
+            elif 'tech_config' in power_sources['pv']:
```
dguittet, line 154: I wonder if there's a more explicit way to let users choose between creating a DetailedPVPlant vs a PVPlant?

hybrid/detailed_pv_plant.py
Matthew-Boyd: I've been running into trouble with the way some properties are set up in HOPP, where you get caught in loops, and that's why I'm setting these directly. In this case, the _actual_ system_capacity has been calculated (without designing a new layout) and should be set in the pysam model as-is, where the property is set up to assume it's been given the _target_ system_capacity and it then calculates a new layout and sets a different system_capacity.

hybrid/hybrid_simulation.py
Matthew-Boyd: I was wondering this too... let's discuss.

hybrid/layout/pv_design_utils.py
```diff
@@ -0,0 +1,316 @@
...
+    modules_per_string: float,
+    module_power: float,
+    inverter_power: float,
+    n_inverters_orig: float,
+    ) -> list:
+    """
+    Ensure coherence between parameters for detailed PV model (pvsamv1),
+    keeping the dc/ac ratio approximately the same
+
+    inputs:
```
dguittet, line 136: please check the formatting of the docstring

dguittet (changes requested):

hybrid/hybrid_simulation.py
```diff
@@ -148,7 +149,12 @@ def __init__(self,
             power_sources[k.lower()] = power_sources.pop(k)
 
         if 'pv' in power_sources.keys():
-            self.pv = PVPlant(self.site, power_sources['pv'])
+            if 'pv_plant' in power_sources['pv']:
+                self.pv = power_sources['pv']['pv_plant']                       # User instantiated plant
+            elif 'use_pvwatts' in power_sources['pv'].keys() and not power_sources['pv']['use_pvwatts']:
+                self.pv = DetailedPVPlant(self.site, power_sources['pv'])       # PVSAMv1 plant
```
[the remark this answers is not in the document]
Matthew-Boyd: Done.

hybrid/detailed_pv_plant.py
```diff
@@ -0,0 +1,97 @@
...
+        return self._system_model.Lifetime.dc_degradation
+
+    @dc_degradation.setter
+    def dc_degradation(self, dc_deg_per_year: Sequence):
+        self._system_model.Lifetime.dc_degradation = dc_deg_per_year
+
+    @property
+    def dc_ac_ratio(self) -> float:
+        return self.system_capacity * 1e3 / \
+            (self.value('inverter_count') * get_inverter_power(self._system_model))
```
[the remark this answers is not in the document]
Matthew-Boyd: I added the resizing capability in pv_design_utils.py:
[https://github.com/NREL/HOPP/blob/312df60f78aa92e5e48f51fc7e1b10afe680c0e1/hybrid/layout/pv_design_utils.py#L55](url)

which I employ in a test here:
[https://github.com/NREL/HOPP/blob/312df60f78aa92e5e48f51fc7e1b10afe680c0e1/tests/hybrid/test_hybrid.py#L175](url)

I don't remember discussing adding properties for these elec-config variables, but I'll add those now.

hybrid/detailed_pv_plant.py
```diff
@@ -0,0 +1,97 @@
...
+        to enforce coherence between attributes
+        """
+        self.assign(params)
+        calculated_system_capacity = verify_capacity_from_electrical_parameters(
+            system_capacity_target=self.value('system_capacity'),
+            n_strings=self.value('subarray1_nstrings'),
+            modules_per_string=self.value('subarray1_modules_per_string'),
+            module_power=get_module_power(self._system_model) * 1e-3
+        )
+        self._system_model.SystemDesign.system_capacity = calculated_system_capacity
```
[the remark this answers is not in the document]
Matthew-Boyd: You're right, now it can after the change to that setter. Changed.

hybrid/layout/pv_design_utils.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+def find_inverter_count(
+    dc_ac_ratio: float,
+    modules_per_string: float,
+    n_strings: float,
+    module_power: float,
+    inverter_power: float,
+    ):
+    """Sizes the number of inverters"""
```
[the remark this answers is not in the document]
Matthew-Boyd: Done

hybrid/layout/pv_design_utils.py
Matthew-Boyd: Fixed

hybrid/detailed_pv_plant.py
```diff
@@ -0,0 +1,147 @@
...
+        """Modules per string"""
+        return self._system_model.SystemDesign.subarray1_modules_per_string
+
+    @modules_per_string.setter
+    def modules_per_string(self, _modules_per_string: float):
+        """Sets the modules per string and updates the system capacity"""
+        self._system_model.SystemDesign.subarray1_modules_per_string = _modules_per_string
+        self._system_model.SystemDesign.subarray1_modules_per_string = 0 
+        self._system_model.SystemDesign.subarray1_modules_per_string = 0
+        self._system_model.SystemDesign.subarray1_modules_per_string = 0
```
dguittet, line 119: Should this be subarray2-4?

hybrid/detailed_pv_plant.py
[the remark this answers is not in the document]
dguittet: Ahh I see thanks

dguittet (approved):
I think this is good to merge after that minor subarray2-4 change!

hybrid/detailed_pv_plant.py
Matthew-Boyd: Agh, darn it! Thanks for the catch!

PR #96, opened 2023-02-08
Run simple PV model PVWattsv8 with a custom detailed layout

Matthew-Boyd:
@dguittet Could you please take a look and ensure this is what Engie needs?

PR #100, opened 2023-02-12
Simple pv detailed layout

dguittet:
Move feature in https://github.com/NREL/HOPP/pull/96 to master


Matthew-Boyd:
@dguittet Do we actually want to merge this into master? I was under the impression that detailed_pv_layout.py (and its test file) are just templates for Engie. (However, find_modules_per_string() and find_strings_per_inverter() could be reused in pv_layout.py). Could you please advise/clarify?

Matthew-Boyd:
If not merging into master, master would be merged into this feature branch, and the other changes in these commits that are not for the detailed PV layout would be included in the other PRs (that are being added asap).

PR #101, opened 2023-02-13
Feature/offshore h2

ckiefer14:
Added to desal_model.py mass (kg) and sizing (m^2).
Added respective tests to test_RO_desal.py.

Desal Limerick from ChatGPT
There once was a water so brine,
That drinking it caused much decline,
But with desalination,
Its taste became elation,
Now thirst is no longer a bane!

(This is a poorly made Limerick as bane and decline do not rhyme but should)


jaredthomas68:
@ckiefer14 this looks pretty good. I think we need some clarification for which mass is equipment mass and which mass is water mass. Can you make that more clear through naming conventions for the equipment?

jaredthomas68:
@ckiefer14 I'd like to see a comparison of sizes. Since you based this on just a single size of desal unit, we're going to be missing the true scaling. Can you check the linear scaling assumption based on available data on a plot?

jaredthomas68:
I'm going ahead with merge, but would still like the previous comments addressed in a future pull request.

PR #109, opened 2023-02-21
Enable custom financial models

Matthew-Boyd:
- Enables custom financial models but does not implement one.
- Includes a test using Singleowner as the custom financial model.

@dguittet The actual custom financial model templates will be added on the other feature branch. At that point, any missing functionality that is discovered can be implemented here in master.

### DIFF
--- a/hybrid/detailed_pv_plant.py
+++ b/hybrid/detailed_pv_plant.py
@@ -22,9 +22,8 @@ class DetailedPVPlant(PowerSource):
 
         :param pv_config: dict, with following keys:
             'tech_config': dict, contains parameters for pvsamv1 technology model
-            'fin_config': dict, contains `model_type` and any inputs for chosen financial model type
+            'fin_model': optional financial model object to use instead of singleowner model
             'layout_params': optional DetailedPVParameters, the design vector w/ values. Required for layout modeling
-            'layout_config': optional dict, contains all keys for PVLayoutConfig dataclass. Required for layout modeling
         """
         system_model = Pvsam.default("FlatPlatePVSingleOwner")
 


hybrid/hybrid_simulation.py
```diff
@@ -179,7 +179,10 @@ def __init__(self,
         if 'geothermal' in power_sources.keys():
             raise NotImplementedError("Geothermal plant not yet implemented")
         if 'grid' in power_sources.keys():
-            self.grid = Grid(self.site, power_sources['grid'])
+            if 'grid_source' in power_sources['grid']:
+                self.grid = power_sources['grid']['grid_source']                # User instantiated grid source
```
dguittet, line 183: What is this "grid_source" option? Can you please add some documentation?

hybrid/power_source.py
```diff
@@ -238,27 +241,35 @@ def simulate_financials(self, interconnect_kw: float, project_life: int):
...
-            single_year_gen = self._financial_model.SystemOutput.gen
-            self._financial_model.SystemOutput.gen = list(single_year_gen) * project_life
-
-        if self.name != "Grid":
-            self._financial_model.SystemOutput.system_pre_curtailment_kwac = self._system_model.value("gen") * project_life
-            self._financial_model.SystemOutput.annual_energy_pre_curtailment_ac = self._system_model.value("annual_energy")
-            # TODO: Should we use the nominal capacity function here?
+        if len(self._financial_model.value('gen')) == self.site.n_timesteps:
+            self._financial_model.value('gen', self._financial_model.value('gen') * project_life)
+        self._financial_model.value('system_pre_curtailment_kwac', self._financial_model.value('gen'))
```
dguittet, line 258: Do these lines no longer require the "Grid" check? In the detailed_pv_opt branch, I still had this check, but perhaps it's no longer necessary?: https://github.com/NREL/HOPP/blob/f44626a7e6bea80b1aa2293d3d07de2a0c3be008/hybrid/power_source.py#L262

hybrid/power_source.py
```diff
@@ -238,27 +241,35 @@ def simulate_financials(self, interconnect_kw: float, project_life: int):
...
+        self._financial_model.value('annual_energy_pre_curtailment_ac', self._system_model.value("annual_energy"))
+        # TODO: Should we use the nominal capacity function here?
         self.gen_max_feasible = self.calc_gen_max_feasible_kwh(interconnect_kw)
         self.capacity_credit_percent = self.calc_capacity_credit_percent(interconnect_kw)
+        if not isinstance(self._financial_model, Singleowner.Singleowner):
+            try:
+                self._financial_model.set_financial_inputs(flatten_dict(self._system_model.export()))
+            except:
+                raise NotImplementedError("Financial model cannot set its inputs.")
+        elif self.name!= "Grid" and isinstance(self._system_model, Pvsamv1.Pvsamv1):
```
dguittet, line 268: I feel like if it's possible, this should be moved to the detailed_pv_plant's `simulate_financials`:

```
super().simulate_financials()
pvsamv1 battery stuff
```

hybrid/hybrid_simulation.py
Matthew-Boyd: It's the instantiated grid object that can be passed to HybridSimulation, implemented the same way as the instantiated 'pv_plant' object a few lines above that. I'll add documentation.

hybrid/power_source.py
Matthew-Boyd: Correct, I refactored this after finding it unnecessary.

hybrid/power_source.py
Matthew-Boyd: You're right, that's much cleaner.

hybrid/detailed_pv_plant.py
```diff
@@ -27,12 +27,15 @@ def __init__(self,
             'layout_config': optional dict, contains all keys for PVLayoutConfig dataclass. Required for layout modeling
         """
         system_model = Pvsam.default("FlatPlatePVSingleOwner")
-        financial_model = Singleowner.from_existing(system_model, "FlatPlatePVSingleOwner")
+
```
dguittet, line 30: Could you please add this option of pv_config to the docstring above?

dguittet (approved):
a last quick documentation request, but then it's good to merge!

hybrid/detailed_pv_plant.py
Matthew-Boyd: Updated, and removed parameter in docstring that will come next with the detailed layout implementation

PR #110, opened 2023-02-22
Fix KML file reader

dguittet:
Add UTM support to KML file reader

SiteInfo reads in the data in the Google Maps KML coordinate projection "EPSG:4326" and converts it to the correct UTM zone projection based on lat and lon. This now results in the correct shape and area.

Next, a PR will be needed to update the KML writer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
         run: |
           sudo apt-get update && sudo apt-get install -y libglpk-dev glpk-utils coinor-cbc
           python -m pip install --upgrade pip
+          pip install scikit-learn
           pip install -r requirements.txt
       - name: Install package
         run: |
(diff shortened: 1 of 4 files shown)


Matthew-Boyd (approved):
I don't see any issues, although I don't have any background with this. I'm hopeful about that CI addition of scikit-learn!

Angus-Mathieson:
Hi, hope you don't mind me commenting! I installed and cloned HOPP this morning but get the error "No module named 'fastkml' " when running any of the examples. I noticed that there has been some work in fixing the KML reader above, but has anything changed in the installation that could mean that it doesn't work for the latest release? Thank you so much!

(Traceback below)

 
image

dguittet:
Hello @Angus-Mathieson, fastkml is in the requirements.txt so it should have been installed when you installed HOPP. Can you check whether it's installed or not?

PR #113, opened 2023-03-03
Feature/green steel ammonia

cfrontin:
Change to using installation-relative path specifications, in order to step towards better tool/project separation.

With these, code should still function as previously, but if, e.g., `HOPP/green_steel_ammonia_run_scenarios.py` were run outside of the HOPP directory (and with appropriate adjustments to imports), the tools from `hopp_tools_steel.py` and other helpers would load successfully.


jaredthomas68:
This looks fine to me. Requesting others reviews from primary authors/users.

PR #114, opened 2023-03-06
Bugfix: Fixing tests and updating CapEx return of the Singlitico PEM cost model

bayc:
**Feature or improvement description**
This PR updates the tests within `tests/hopp` to working versions, and updates the CapEx Singlitico cost model to return total cost for the system, not on a cost per unit basis.

**Related issue, if one exists**
None.

**Impacted areas of the software**
`tests/hopp`
`hopp/hydrogen/electrolysis/PEM_costs_Singlitico_model.py`

**Additional supporting information**
I did have to change a PTC regression test value in `test_hybrid.py` that I am not sure what had changed or if it was out of date (the value was [here](https://github.com/NREL/HOPP/pull/114/files#diff-dd621a8aed689a8d862dc9ac9bc2e3386e63708e799ec3682649e771f3fa7ed3R286)). It did not change by much.

Also, in `test_layout.py`, I had to update the expected wind turbine coordinates. My best guess for this is that the input for the tests changed, resulting in different locations of the turbines from before, but would be good to get a second set of eyes on it.

The Shapely related errors were caused by Shapely's change of its `MultiPoint` object, starting with v2.0. The `MultiPoint` object is no longer an iterable, and to access the values/tuples of the object, you need to use `.geoms`, as noted in Shapely's change-log.

**Test results, if applicable**
All tests in `tests/hopp` should now pass.

### DIFF
--- a/hopp/hydrogen/electrolysis/PEM_costs_Singlitico_model.py
+++ b/hopp/hydrogen/electrolysis/PEM_costs_Singlitico_model.py
@@ -101,11 +101,17 @@ class PEMCostsSingliticoModel():
         # If electrolyzer capacity is >100MW, fix unit cost to 100MW electrolyzer as economies of scale
         # stop at sizes above this, according to assumption in [1].
         if P_elec > 100 / 10**3:
-            P_elec = 0.1
+            P_elec_cost_per_unit_calc = 0.1
+        else:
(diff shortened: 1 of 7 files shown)


jaredthomas68:
Looks good. Thanks.

PR #117, opened 2023-03-20
provide grid connected and grid disconnected and hybrid energy profiles

jaredthomas68:
Let me know if this is what you need.


ereznicek (approved):
Looks good

PR #118, opened 2023-03-23
Add capability for custom financial models

Matthew-Boyd:
Adds capability for custom financial models, along with a simple example model for NPV and associated tests demonstrating its usage with PV + wind hybrid models.


dguittet:
@Matthew-Boyd This looks good! One reservation I have is that there are the DataClasses for Depreciation and Tax Incentives that aren't used at all in the custom financial model. I don't think the financial model should process those parts as simple is better. But I think it would be better to remove those unused data from the financial model, and add checks within the HybridSimulation code to enable financial models that don't have those data. So there are parts of financial model code that expects Singleowner, and that code should be modified to be more flexible.

hybrid/financial/custom_financial_model.py
```diff
@@ -0,0 +1,383 @@
...
+        Computes the annual O&M cost from the fixed, per capacity and per production costs
+        """
+        return self.value('om_fixed')[0] \
+               + self.value('om_capacity')[0] * self.value('system_capacity') \
+               + self.value('om_production')[0] * self.value('annual_energy') * 1e-3
+
+
+    def value(self, var_name, var_value=None):
+        # TODO: is this the best way to replicate this functionality within the PySAM value() function,
+        # where (array([15.]),) becomes (15,), like in set_average_for_hybrid("om_capacity", size_ratios) ?
```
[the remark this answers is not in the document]
Matthew-Boyd: Thanks! That does fix the issue.

hybrid/financial/custom_financial_model.py
```diff
@@ -0,0 +1,383 @@
...
+        if (isinstance(var_value, tuple) or isinstance(var_value, list)) and isinstance(var_value[0], np.ndarray):
+            var_value = (float(var_value[0][0]),)
+        elif isinstance(var_value, np.ndarray) and isinstance(var_value[0], np.float64):
+            var_value = float(var_value[0])  # used for aggregate NPV value
```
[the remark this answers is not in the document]
Matthew-Boyd: Done

hybrid/dispatch/power_sources/power_source_dispatch.py
```diff
@@ -67,6 +69,8 @@ def initialize_parameters(self):
     def update_time_series_parameters(self, start_time: int):
         n_horizon = len(self.blocks.index_set())
         generation = self._system_model.value("gen")
+        if isinstance(self._system_model, Pvsamv1.Pvsamv1):
+            generation = tuple(np.clip(generation, 0, None))   # remove negative generation (i.e., load)
```
Matthew-Boyd, line 73: @dguittet I had to do this here as the _system_model is not a detailed_pv_plant here but rather the pvsamv1 pysam model

hybrid/financial/custom_financial_model.py
```diff
@@ -246,8 +193,8 @@ def execute(self, n=0):
                 rate=self.nominal_discount_rate(
                     inflation_rate=self.value('inflation_rate'),
```
dguittet, line 194: is inflation rate supposed to be a ratio or percent?

hybrid/dispatch/power_sources/power_source_dispatch.py
Matthew-Boyd: I over-rode it in `PvDispatch`

hybrid/financial/custom_financial_model.py
Matthew-Boyd: The inflation_rate parameter is a percent in SAM, so nominal_discount_rate is made a percent for consistency with that 'convention'. However, the npv function borrowed from the numpy-financial package takes its rate parameter as a fraction, hence the divide by 100 conversion.

dguittet: is there documentation on the units of the parameters anywhere?

hybrid/dispatch/power_sources/pv_dispatch.py
```diff
@@ -22,3 +22,6 @@ def __init__(self,
                  block_set_name: str = 'pv'):
         super().__init__(pyomo_model, indexed_set, system_model, financial_model, block_set_name=block_set_name)
 
+    def update_time_series_parameters(self, start_time: int):
+        self._system_model.value("gen", [max(0, i) for i in self._system_model.value("gen")])  # zero out any load
+        super().update_time_series_parameters(start_time)
```
dguittet, line 27: I'm not sure we want to overwrite the gen in the system model. I feel like we still want to run the inverter nighttime load as a grid purchase in the financial model

hybrid/dispatch/power_sources/pv_dispatch.py
```diff
@@ -23,5 +23,5 @@ def __init__(self,
         super().__init__(pyomo_model, indexed_set, system_model, financial_model, block_set_name=block_set_name)
 
     def update_time_series_parameters(self, start_time: int):
-        self._system_model.value("gen", [max(0, i) for i in self._system_model.value("gen")])  # zero out any load
         super().update_time_series_parameters(start_time)
+        self.available_generation = [max(0, i) for i in self.available_generation]  # zero out any load
```
dguittet, line 27: in the comment I think you meant "zero out any negative load"

PR #119, opened 2023-03-26
Flicker for rotated arrays

dguittet:
Enables calculation of flicker losses for arrays that are rotated.

Allow `get_flicker_loss_multiplier` to use `strands` or `module_points`, shapely MultiPoint of module locations. Strands are LineStrings where the modules are located in a single row. Neither strands nor `module_points` differentiate between strings or racks.

Tested in test_layout/test_hybrid_layout_rotated_array. The runtime using MultiPoint is slightly faster (6% faster in a single trial).

image


dguittet:
Thanks @Matthew-Boyd

PR #120, opened 2023-03-27
Distributed pressure vessel computations

cfrontin:
Added distributed pressure vessel functions.

Additive changes only.

All new methods unit tested against ~~gold~~ pyrite values.


jaredthomas68 (commented):
This looks pretty good. 

Note that energy is in kWh/kg, so the energy should not be multiplied by `N_sites` in the distributed case as I understand it.

Also, it would be good to check the monolithic and distributed results against each other relatively. This is especially important since we don't have any hard values to test against. At least we know the monolithic should be cheaper, use less energy, etc.

cfrontin:
> Note that energy is in kWh/kg, so the energy should not be multiplied by N_sites in the distributed case as I understand it.

I'm not sure that that's true? See: https://github.com/NREL/HOPP/blob/7a4055da44863503c03b9f534913252095a1cd78/hopp/hydrogen/h2_storage/pressure_vessel/compressed_gas_storage_model_20221021/Compressed_all.py#L73

> Also, it would be good to check the monolithic and distributed results against each other relatively.

For this you mean just a `>=` comparison test? Or perhaps making sure a ratio matches?

PR #122, opened 2023-03-31
added changes to make pressure vessel storage at-turbine work.

cfrontin:
This definitely needs to be checked out!

There is not a lot of sanity-checking on how the `design_scenario` and `plant_config` settings collide in hopp/eco/hydrogen_mgmt.py- current settings run `scenario8`


jaredthomas68 (changes requested):
Looks like it's heading the right direction. The only concern I have is that the pressure vessel location (h2_storage) is now conflated with the electrolyzer location (h2_location). This is my fault for introducing poor naming conventions, but we need to fix it and also fix the logic errors it introduced. Let's change "h2_location" to "electrolyzer_location" and "h2_storage" to "h2_storage_location" and then correct the remaining logic errors.

cfrontin (commented):
I think this is getting here. Still working through to make sure the logic works.

PR #123, opened 2023-04-06
Interim for merge

ereznicek:
Merging in Elenya's work that added degradation modeling to the electrolyzer and creating an option to perform a parametric sweep on solar and battery sizes.

Jared - can you make sure that the changes to hopp_tools_steel.grid are ok?

Sanjana - can you make sure that none of the changes will break your optimization routine?

### DIFF
--- a/hopp_tools_steel.py
+++ b/hopp_tools_steel.py
@@ -1036,6 +1036,9 @@ def grid(
                 cost_to_buy_from_grid += energy_from_the_grid[i]*buy_price
                 energy_total[i] = energy_from_the_grid[i] + combined_pv_wind_storage_power_production_hopp[i]
                 energy_to_electrolyzer[i] = kw_continuous
+            else:
+                energy_to_electrolyzer[i] = kw_continuous
+                energy_total[i]=kw_continuous
     else:
         cost_to_buy_from_grid = 0.0
         energy_to_electrolyzer = [x if x < kw_continuous else kw_continuous for x in combined_pv_wind_storage_power_production_hopp]
--- a/hopp_tools_steel.py
+++ b/hopp_tools_steel.py
@@ -1473,7 +1473,9 @@ def write_outputs_RODeO(electrical_generation_timeseries,
            wind_itc_total, total_itc_hvdc\
 
 def write_outputs_ProFAST(electrical_generation_timeseries,
-                         hybrid_plant,
+                         cf_wind_annuals,
+                         cf_solar_annuals,
+                         wind_itc_total,
                          total_export_system_cost,
                          total_export_om_cost,
                          cost_to_buy_from_grid,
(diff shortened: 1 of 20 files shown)


hopp_tools_steel.py
```diff
@@ -2199,11 +2306,11 @@ def hydrogen_storage_capacity_cost_calcs(H2_Results,electrolyzer_size_mw,storage
     
     # Get average electrolyzer efficiency
     electrolyzer_efficiency_while_running = []
-    for j in range(len(H2_Results['electrolyzer_total_efficiency'])):
+    for j in range(len(H2_Results['water_hourly_usage'])):
         if H2_Results['hydrogen_hourly_production'][j] > 0:
             electrolyzer_efficiency_while_running.append(H2_Results['electrolyzer_total_efficiency'][j])
```
jaredthomas68, line 2311: This looks like an incomplete loop/if statement. Also not sure why the water hourly usage is being used to loop over the hydrogen hourly production. Loops should be over the relevant iteratable. We need to make sure everything runs before merging. This should fail as is.

hopp_tools_steel.py
```diff
@@ -555,39 +562,58 @@ def run_HOPP(
...
     rotor_diameter = scenario['Rotor Diameter']
     
     if floris == False:
-        if storage_size_mw > 0:
-            technologies = {#'pv':
-                            #   {'system_capacity_kw': solar_size_mw * 1000},
-                            'wind':
-                                {'num_turbines': np.floor(wind_size_mw / turbine_rating_mw),
+        technologies={}
+        if run_wind_plant:
```
jaredthomas68, line 577: Instead of being a flag, `run_wind_plant` checks should be replaced with a check on the wind size as is done with the `solar_size_mw` to keep things consistent and allow similar usage.

jaredthomas68:
We would ideally have some tests in place for the system to make sure the two system codes provide the same results. Have you checked that you get the same outputs for the same outputs across both branches?

jaredthomas68 (changes requested):

hopp_tools_steel.py
ereznicek: I agree. It seems to work without issue as it is here so I suggest we leave it for now, but note that in the future we should replace it with a more accurate calculation of average efficiency as you noted in point # 2 above.

hopp_tools_steel.py
elenya-grant: Is this change recommended for this function only or all cases where we use run_wind_plant? If it's just in this function, then that can be done easily. If the change is suggested for more than that, it gets more complicated.

If just in this function: This would require setting wind_size_mw=0 in the main run script (or making a new variable that we set to the actual installed wind capacity when running the wind plant and setting to zero when not running the wind plant). If we set wind_size_mw=0 in the run file, although we are using wind power just not running the wind farm model, then it could have some downstream impacts and cause issues (notably in the lcoh calculation). My idea would be to go the route of using a new variable, which would be set depending on the run_wind_plant flag in the run script.

To remove the run_wind_plant flag in the main run script and keep the same parametric sweep functionality, then I think we'd need to introduce a variable called wind_sizes_mw which is a list of length = len(solar_sizes_mw)*len(storage_sizes_mw), the first value is the actual wind capacity, all the rest are zero. That way, wind_size_mw is always the installed wind capacity (hopefully preventing any potential downstream issues), but the run_wind_plant flag is replaced with "if wind_sizes_mw[j]>0" and pass in wind_sizes_mw[j] to the run_HOPP function (that way it works with the change I suggested previously).

hybrid/PEM_Model_2Push/run_PEM_master.py
```diff
@@ -0,0 +1,208 @@
...
+        #Do not modify stack_rating_kw or stack_min_power_kw
+        #these represent the hard-coded and unmodifiable 
+        #PEM model basecode
+        self.stack_rating_kw = 1000 #single stack rating - DO NOT CHANGE
+        self.stack_min_power_kw=0.1*self.stack_rating_kw
+        self.input_power_kw=electrical_power_signal
+        self.cluster_min_power = self.stack_min_power_kw * self.cluster_cap_mw
+        self.cluster_max_power = self.stack_rating_kw * self.cluster_cap_mw
+
+
```
vijay092, line 71: would be good to keep track of the sampling time. `self.dt_hr = 1`

vijay092 (commented):
I only looked at the parts of the code that will affect my optimization algorithm and LGTM overall!

PR #124, opened 2023-04-13
Add electrolyzer optimization module

vijay092:
This PR introduces an optimization module designed to maximize H2 production while minimizing degradation. These results are based on a one-year timeframe. 

* The addition of the degradation constraint has little impact on the H2 production.
image

* However, it significantly reduces the switching cost, estimated at $12 per transition.
image

* Some stats in the computational time: 
image

Notes:
* To use the `cbc` solver, please run the following command: `conda install -c conda-forge pyomo coincbc`.
* Details of the formulation can be found here: https://sharelatex.nrel.gov/read/sdnngsdtcrrj


ereznicek (approved):
Looks good. Can't say from looking at it if it will run correctly or not but it shouldn't mess up any of the existing code base.

PR #125, opened 2023-04-18
Calculate number of combiner boxes during electrical sizing

Matthew-Boyd:
Addresses request from Engie via email on 4/18.


dguittet (approved):
Thanks for the quick turnaround

PR #130, opened 2023-05-11
Add pvsamv1 properties and setters for interdependent parameters

Matthew-Boyd:
This includes properties and setters for all interdependent parameters, including:

- system capacity
- modules per string
- number of strings (restricting to subarray1 only)
- PV module model
- inverter model


dguittet (commented):
Another thought I had was that it would be good to move a lot of the functionality in `pv_design_utils.py` and `pv_inverter`.py` into PySAM where it'll probably get a broader user base. Not something we can do with the current project, but I'll ask if it's appropriate for SAM core

dguittet:
@Matthew-Boyd I added an assert for the PV system capacity in here: https://github.com/NREL/HOPP/pull/130/commits/a6275ccdf2b333ca84bf607a0bddfeb543e2b60f

This now causes the `test_hybrid_detailed_pv_only` to fail.

dguittet:
@Matthew-Boyd Looking at the latest change, I think it went in the wrong direction. The DetailedPVPlant needs to be able to use a system capacity. 

Can you tell me what the input dictionary format for DetailedPVPlant should be? Where does the system_capacity input go?

IIRC, the input dictionary format for the simple PVPlant is different since there's no "tech_config" required.

Matthew-Boyd:
> @Matthew-Boyd Looking at the latest change, I think it went in the wrong direction. The DetailedPVPlant needs to be able to use a system capacity.
> 
> Can you tell me what the input dictionary format for DetailedPVPlant should be? Where does the system_capacity input go?
> 
> IIRC, the input dictionary format for the simple PVPlant is different since there's no "tech_config" required.

@dguittet 
The dictionary key `system_capacity_kw` in the top level of the `pv_config` parameter for `DetailedPVPlant` was not being used before this last commit. The value for system capacity was only being taken from the `tech_config` sub-dict in `pv_config`.

The input dictionary format for the `DetailedPVPlant`, as provided in `tech_config`, is demonstrated in `pvsamv1_basic_params.json`, which is the same as the SAM UI code generation for "PySAM JSON". The system capacity is adjusted for coherence with other model parameters and set in the model in `DetailedPVPlant::processed_assign()`

I will add another test to ensure the `system_capacity_kw` top-level key value is used properly and in place of any value for `system_capacity` in `tech_config`.

PR #132, opened 2023-05-16
Pysam update to 4.1.0

dguittet:
Will delay this to wait for release of PySAM 4.2.0, which will include improvements to the solar position algorithm to speed up PV runs by ~2x

@nimalgui @Matthew-Boyd


camirmas:
It appears that 4.2.0 is now available https://github.com/NREL/pysam/releases/tag/v4.2.0

PR #135, opened 2023-07-13
Adding heuristic load following dispatch method and load following ex…

genevievestarke:
Adding heuristic load following dispatch method in power_storage
Added simulate_hybrid_load_following.py as an example of how to run the load following heuristic method
Added load signal to battery generation plots

- [x] Add test: test_hybrid_dispatch_baseload_heuristic_and_analysis and test_hybrid_dispatch_one_cycle_heuristic_baseload from https://github.com/bayc/HOPP-1/blob/archive/new_batt_heuristic/tests/hybrid/test_dispatch.py
- [x] Add baseload options to hybrid_dispatch_options dict: https://github.com/bayc/HOPP-1/blob/archive/new_batt_heuristic/hybrid/dispatch/hybrid_dispatch_options.py#L55
- [x] Add analysis code to grid.py from: https://github.com/bayc/HOPP-1/blob/archive/new_batt_heuristic/hybrid/grid.py#L81

### DIFF
--- a/examples/simulate_hybrid_load_following.py
+++ b/examples/simulate_hybrid_load_following.py
@@ -1,0 +1,94 @@
+from pathlib import Path
+from hopp.simulation.technologies.sites import SiteInfo, flatirons_site
+from hopp.simulation.hybrid_simulation import HybridSimulation
+from hopp.simulation.technologies.dispatch.plot_tools import plot_battery_output, plot_battery_dispatch_error, plot_generation_profile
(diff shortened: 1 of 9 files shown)


genevievestarke:
@bayc I added the analysis code (updated a little based on heuristic method) and the test, and updated the example script to use this functionality. It's running for me, so let me know if you have trouble with it!

camirmas:
closing, #262 uses this code but updates it for HOPP v2.0+

PR #195, opened 2023-08-17
Update Custom Financial Model

dguittet:
Fixes some bugs in how data from CustomFinancialModel is passed back and forth with HybridSimulation, making sure that the system's values are consistent throughout.

The `power_sources` that are able to be paired with a custom financial model include: `battery`, `detailed_pv_plant`, `power_source`, `pv_source`, `wind_source`. The ones not currently enabled are `csp_source` and `trough_source` due to how these models use ssc directly and I am not sure what adjustments are needed

CustomFinancialModel (and any other custom ones in the future) needs to have these functions in order to be integrated with `power_source`, which checks for these in `import_financial_model`
- `set_financial_model`
- `assign`
- `value`
- `unassign`
- `execute`

During `simulate_financials`, the values of the system's power simulation are passed to the financial model. 

Some minor changes for catching warnings, logging, etc.

Also modify `power_source`'s `calc_capacity_credit_percent` so that it can be computed either in the existing way of "cap_hours" or in the new way of using an external capacity credit input.


dguittet:
@qualand Thanks for the review!!

PR #196, opened 2023-08-18
BatteryStateless

dguittet:
Make a"stateless" alternative to the Battery class. The Battery class uses `PySAM.BatteryStateful` module to simulate the battery using the dispatch power, resulting in potentially different power output than the dispatch. BatteryStateless does not have this module, instead it pulls all battery variables such as power and SOC from the dispatch model itself.

Other changes:
 - Remove unneeded battery variables from `CustomFinancialModel`
 - Modify `battery.py` so it has configurable max, min and initial SOC. Also organize battery variables
 - `detailed_pv_plant.py`: after cleaning up the battery variables, this class no longer needs to derive the `simulate_financials` class
 - `hybrid_dispatch`: refactor the objective for reporting, make `time_weighting_factor` configurable
 - add `test_battery.py` to check that the dispatch of Battery and BatteryStateless are the same

**needs to be merged after the custom financial model updates PR are merged** #195


hybrid/battery_stateless.py
```diff
@@ -0,0 +1,213 @@
...
+    @property
+    def SOC(self) -> float:
+        if len(self.Outputs.SOC):
+            return self.Outputs.SOC[0]
+        else:
+            return self.initial_SOC
+
+    @property
+    def lifecycles(self) -> float:
+        return self.dispatch
```
qualand, line 213: Should this be `return self.Outputs.lifecycles_per_day` ?

qualand (approved):
@dguittet see two minor comments

hybrid/battery_stateless.py
dguittet: Yes, thanks!

hybrid/csp_source.py
```diff
@@ -178,7 +178,7 @@ def tmy3_to_df(self):
         df.index.name = 'datetime'
         df.drop(date_cols, axis=1, inplace=True)
 
-        df.index = df.index.map(lambda t: t.replace(year=df.index[0].year))  # normalize all years to that of 1/1
+        # df.index = df.index.map(lambda t: t.replace(year=df.index[0].year))  # normalize all years to that of 1/1
```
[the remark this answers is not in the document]
dguittet: I had commented this out because it caused an issue in more recent versions of pandas. I'm uncommenting it out now, but I think I need to pin the version of pandas

PR #201, opened 2023-08-22
Site info refactor

camirmas:


camirmas:
Something isn't quite right here, I think I used an outdated branch. Going to open a new one

PR #202, opened 2023-08-22
Refactor SiteInfo

### LINKED ISSUE #145: Refactor SiteInfo class to use attrs dataclass
Refactor the `SiteInfo` class in `site_info.py` to use the attrs library dataclass and clean up the `__init__` function.

camirmas:
# Refactor site_info to use the attrs library and method of class definition

This pull request refactors the `SiteInfo` class to use the attrs library dataclass functionality. This cleans up the `__init__` function and doesn't require the user to write some boilerplate code for each class. It also adds the ability to define converters and validators for input, building robustness into the code. Basic unit tests are added for `SiteInfo` and the docstrings involved have been updated to use Google Sphinx [format](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html).

## Related issue
resolves #145 

Mainly `site_info.py`, with updates to wind/solar resource inputs as needed elsewhere.

None

Build passing [here](https://github.com/camirmas/HOPP/actions/runs/5943969928)

### DIFF
--- a/hopp/simulation/technologies/sites/site_info.py
+++ b/hopp/simulation/technologies/sites/site_info.py
@@ -1,19 +1,31 @@
+from typing import Optional, Union
+from pathlib import Path
+
+from attrs import define, field
 import matplotlib.pyplot as plt
-from shapely.geometry import *
-from shapely.geometry.base import *
-from shapely.validation import make_valid
-from fastkml import kml
+import numpy as np
+from numpy.typing import NDArray
+from shapely.geometry import Polygon, MultiPolygon, Point 
+from shapely.geometry.base import BaseGeometry
 from shapely.ops import transform
+from shapely.validation import make_valid
+from fastkml import kml, KML
 import pyproj
 import utm
 
-from hopp.simulation.technologies.resource.solar_resource import SolarResource
-from hopp.simulation.technologies.resource.wind_resource import WindResource
-from hopp.simulation.technologies.resource.elec_prices import ElectricityPrices
+from hopp.simulation.technologies.resource import (
+    SolarResource,
+    WindResource,
+    ElectricityPrices
+)
 from hopp.simulation.technologies.layout.plot_tools import plot_shape
 from hopp.utilities.log import hybrid_logger as logger
 from hopp.utilities.keys import set_nrel_key_dot_env
-
+from hopp.type_dec import (
+    hopp_array_converter as converter, NDArrayFloat, resource_file_converter,
+    hopp_float_type
+)
+from hopp.simulation.base import BaseClass
 
 def plot_site(verts, plt_style, labels):
     for i in range(len(verts)):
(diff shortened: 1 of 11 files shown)


camirmas:
Build passing [here](https://github.com/camirmas/HOPP/actions/runs/5943969928)

hopp/type_dec.py
```diff
@@ -35,17 +35,20 @@
...
 ### Custom callables for attrs objects and functions
 
-def hopp_array_converter(data: Iterable) -> np.ndarray:
-    try:
-        a = np.array(data, dtype=hopp_float_type)
-    except TypeError as e:
-        raise TypeError(e.args[0] + f". Data given: {data}")
-    return a
+def hopp_array_converter(dtype: Any = hopp_float_type) -> Callable:
+    def converter(data: Iterable):
```
bayc, line 39: What is the reason to return a converter?

hopp/tools/dispatch/csp_pv_battery_plot.py
```diff
@@ -84,7 +83,8 @@ def init_hybrid_plant():
     site = SiteInfo(site_data,
```
bayc, line 83: This may not be the time to do it, but I would prefer we move to the convention that function calls that need to be broken get moved to just 1 indent, like below. Also, to add the trailing comma on the last input. But this could be handled with commit hooks.

```
site = SiteInfo(
    site_data,
    solar_resource_file=solar_file,
    grid_resource_file=prices_file,
    desired_schedule=desired_schedule
    desired_schedule=desired_schedule,
    wind=False,
)
```

hopp/simulation/technologies/sites/site_info.py
```diff
@@ -1,19 +1,31 @@
...
+    ElectricityPrices
+)
 from hopp.simulation.technologies.layout.plot_tools import plot_shape
 from hopp.utilities.log import hybrid_logger as logger
 from hopp.utilities.keys import set_nrel_key_dot_env
-
+from hopp.type_dec import (
+    hopp_array_converter as converter, NDArrayFloat, resource_file_converter,
+    hopp_float_type
+)
```
bayc, line 27: Would prefer this broken out by line:

```
from hopp.type_dec import (
    hopp_array_converter as converter,
    NDArrayFloat,
    resource_file_converter,
    hopp_float_type,
)
```

hopp/simulation/technologies/sites/site_info.py
```diff
@@ -25,140 +37,121 @@ def plot_site(verts, plt_style, labels):
...
+    Args:
+        data (dict): Dictionary containing site-specific information.
+        solar_resource_file (Union[Path, str], optional): Path to solar resource file. Defaults to "".
+        wind_resource_file (Union[Path, str], optional): Path to wind resource file. Defaults to "".
+        grid_resource_file (Union[Path, str], optional): Path to grid pricing data file. Defaults to "".
+        hub_height (float, optional): Turbine hub height for resource download in meters. Defaults to 97.0.
+        capacity_hours (:obj:`NDArray`, optional): Boolean list indicating hours for capacity payments. Defaults to [].
+        desired_schedule (:obj:`NDArray`, optional): Absolute desired load profile in MWe. Defaults to [].
+        solar (bool, optional): Whether to set solar data for this site. Defaults to True.
+        wind (bool, optional): Whether to set wind data for this site. Defaults to True.
```
bayc, line 54: Really like this format!

hopp/simulation/technologies/sites/site_info.py
```diff
@@ -25,140 +37,121 @@ def plot_site(verts, plt_style, labels):
...
     """
+    # User provided
+    data: dict
+    solar_resource_file: Union[Path, str] = field(default="", converter=resource_file_converter)
+    wind_resource_file: Union[Path, str] = field(default="", converter=resource_file_converter)
+    grid_resource_file: Union[Path, str] = field(default="", converter=resource_file_converter)
+    hub_height: hopp_float_type = hopp_float_type(97.)
```
bayc, line 61: Why wouldn't this be like below?

```
hub_height: hopp_float_type = field(default=97., converter=hopp_float_type)
```

hopp/simulation/technologies/sites/site_info.py
```diff
@@ -25,140 +37,121 @@ def plot_site(verts, plt_style, labels):
...
+    capacity_hours: NDArray = field(default=[], converter=converter(bool))
+    desired_schedule: NDArrayFloat = field(default=[], converter=converter())
+    solar: bool = True
+    wind: bool = True
```
bayc, line 65: Should `solar` and `wind` have `field` as well?

bayc:
@camirmas I updated the PR description, leaving your original text commented out. This is the format we use on the FLORIS repo and I think it is really useful to have folks add relevant information to their PRs. I will open a PR with templates for PRs, issues, and discussions, but wanted to try it here since this will be an example PR for folks. Interested to hear your thoughts!

hopp/simulation/technologies/sites/site_info.py
camirmas: I like the Google format too, and the Sphinx doc build is already set up to parse them

hopp/simulation/technologies/sites/site_info.py
camirmas: It's not required, I left it simple because I couldn't think of a reason we'd need a validator/converter on there, but I'm fine to add it back if you prefer

camirmas:
@bayc I think the PR template looks good if you'd like to move forward with that. I also just pushed a commit trying out the autoformatter on _only_ the changed files, with the idea that we'd run it each time we refactor some portion of the codebase. I suspect that it will be easier to format in chunks than to do the whole codebase at once, but wanted to get your take.

Here's the command that I ran to autoformat only this PR's changes:

`git diff HEAD~2 --name-only | grep '\.py$' | xargs black`

hopp/type_dec.py
camirmas: Forgot to add one before, but just added a docstring. It's a function that returns a converter for any NumPy `dtype` so we don't have to make new converter functions

hopp/simulation/technologies/sites/site_info.py
bayc: I think it would be nice to include the field method, as these will eventually be raised to the input file level. Then can still be optional.

camirmas:
@bayc Following up from our discussion, I just set the head back to before the autoformatter, and will open a separate PR for it.

hopp/simulation/technologies/sites/site_info.py
camirmas: Will be fixed with autoformatting PR

hopp/tools/dispatch/csp_pv_battery_plot.py
camirmas: Will be fixed with autoformatting PR

PR #203, opened 2023-08-23
Battery max cycles

dguittet:
Add the ability to limit how many full-energy cycles per day the battery can operate, and export the number of cycles as Outputs for the user.

Battery.py:
 - switch the BatteryOutputs to a `dataclass`, functionality is the same. Add documentation
 - Add reporting for cycles per day

BatteryStateless.py:
 - Add cycle reporting and documentation

Change API to battery dispatch models initializers so to pass in the `HybridDispatchOptions` instead of just the `include_lifecycle_count` option

Update `_lifecycle_count_rule` functions so that daily cycle limits are respected.

Add new test to test_battery.py that shows how to use this new dispatch option, and update other tests


hybrid/battery.py
```diff
@@ -7,17 +8,52 @@
 from hybrid.power_source import *
 
 
+@dataclass
```
camirmas, line 11: seems like a good use for `dataclass` 👍

hybrid/dispatch/power_storage/linear_voltage_convex_battery_dispatch.py
```diff
@@ -19,14 +19,14 @@ def __init__(self,
                  system_model: BatteryModel.BatteryStateful,
                  financial_model: Singleowner.Singleowner,
                  block_set_name: str = 'convex_LV_battery',
-                 include_lifecycle_count: bool = True,
+                 dispatch_options: dict = {},
```
camirmas, line 22: Setting mutable defaults for arguments is dangerous in Python, I'd recommend defaulting to None and initializing inside the method. For example:

```python
def __init__(self, ..., dispatch_options=None):
    if dispatch_options is None:
        dispatch_options = {}
    ...
```

hybrid/dispatch/power_storage/linear_voltage_nonconvex_battery_dispatch.py
```diff
@@ -19,7 +19,7 @@ def __init__(self,
                  system_model: BatteryModel.BatteryStateful,
                  financial_model: Singleowner.Singleowner,
                  block_set_name: str = 'LV_battery',
-                 include_lifecycle_count: bool = True,
+                 dispatch_options: dict = {},
```
camirmas, line 22: same as my previous comment about mutable defaults

hybrid/dispatch/power_storage/one_cycle_battery_dispatch_heuristic.py
```diff
@@ -19,13 +19,13 @@ def __init__(self,
                  system_model: BatteryModel.BatteryStateful,
                  financial_model: Singleowner.Singleowner,
                  block_set_name: str = 'one_cycle_heuristic_battery',
-                 include_lifecycle_count: bool = False):
+                 dispatch_options: dict = {}):
```
camirmas, line 22: same as my previous comment about mutable defaults

hybrid/dispatch/power_storage/simple_battery_dispatch_heuristic.py
```diff
@@ -19,7 +19,7 @@ def __init__(self,
                  financial_model: Singleowner.Singleowner,
                  fixed_dispatch: list = None,
                  block_set_name: str = 'heuristic_battery',
-                 include_lifecycle_count: bool = False):
+                 dispatch_options: dict = {}):
```
camirmas, line 22: same as my previous comment about mutable defaults

camirmas (commented):
Looks good, my only concern is with the mutable default arguments

hybrid/battery.py
dguittet: thanks!

hybrid/dispatch/power_storage/linear_voltage_convex_battery_dispatch.py
dguittet: Thanks fixed

camirmas (approved):

PR #205, opened 2023-08-24
Remove network calls from tests

### LINKED ISSUE #145: Refactor SiteInfo class to use attrs dataclass

### LINKED ISSUE #152: Remove resource network calls from tests
Currently, some of the tests require resource downloads using `NREL_API_KEY`. Stubbing out these network calls in tests will improve performance and stability.

- Identify relevant network calls in tests
- Download relevant resources and save them somewhere accessible in `tests`
- Stub out any methods that make resource calls, instead returning those saved results
- Remove the requirement for `NREL_API_KEY` when we're in a testing environment

An added benefit of this change---and one that I think warrants making this a higher priority---is that PRs made from forks will be able to see tests pass. This is not currently possible because `NREL_API_KEY` cannot be used on any job triggered on the main repo via a fork, so all tests requiring the API key will fail, even if they pass on the fork (where the user has supplied their own `NREL_API_KEY`).

camirmas:
# Remove network calls from tests

This update removes network calls wherever they exist, and also adds new unit tests for functionality related to API calls. It achieves this by mocking every request made using `requests` using the `responses` package. Mocking at the HTTP response level lets us ensure that HOPP is still creating/handling API requests properly during testing, without actually making network calls.

- Minimal changes to source files, mostly related to exposing the API base url so that it can be imported for testing
- Added a new (development only) dependency, `resources`: https://github.com/getsentry/responses. It's developed by Sentry, and simplifies stubbing out network calls. I added a new test directory, `api_responses`, which contains YAML files representing these HTTP responses
- Re-added a solar resource file. I previously deleted this file, but it's actually needed by `test_resource_download.py`. Until now, that test would make a network call for the desired file
- Adds a new `pytest` configuration file, `conftest.py`, that sets a dummy `NREL_API_KEY`
- CI builds no longer set `NREL_API_KEY` as a build step
- Added new tests for `Resource` class

## Related issue
resolves #145 

Primarily tests

None

Build passing [here](https://github.com/camirmas/HOPP/actions/runs/5967713945/job/16189956157) but we'll need to see them pass on the PR before merging this one.


camirmas:
Tests passing [here](https://github.com/camirmas/HOPP/actions/runs/5967713945/job/16189956157) but we should see them pass on the PR once it's picked up off the Queue

hopp/simulation/technologies/layout/flicker_mismatch.py
```diff
@@ -167,7 +169,7 @@ def _setup_irradiance(self):
         pv_model.SystemDesign.gcr = .1
         if self.solar_resource_data is None:
             filename = str(self.lat) + "_" + str(self.lon) + "_psmv3_60_2012.csv"
-            weather_path = Path(__file__).parent.parent.parent / "resource_files" / "solar" / filename
+            weather_path = ROOT_DIR.parent / "resource_files" / "solar" / filename
```
bayc, line 172: I like this! Will help clean up a lot of code.

bayc (approved):

Issue #207, opened 2023-08-25
maximum_power setter not called


jaredthomas68:
When I try to run `examples/optimization/hybrid_sizing_problem.py` as in the draft pull request from jaredthomas68/branch:setter_bug, it appears that the setter for `maximum_power` in the `PowerStorageDispatch` class is not being called. I tried making sure the `PowerStorageDispatch` class was inheriting from `object` ultimately, but that did not seem to help.

`Traceback (most recent call last):
  File "hybrid_sizing_problem.py", line 72, in <module>
    hybrid_plant.battery.dispatch.set_fixed_dispatch(fixed_dispatch, grid_limit=np.ones(len(fixed_dispatch)))
  File "/HOPP/hopp/simulation/technologies/dispatch/power_storage/simple_battery_dispatch_heuristic.py", line 50, in set_fixed_dispatch
    self._set_power_fraction_limits(gen, grid_limit)
  File "/HOPP/hopp/simulation/technologies/dispatch/power_storage/simple_battery_dispatch_heuristic.py", line 67, in _set_power_fraction_limits
    self.max_charge_fraction[t] = self.enforce_power_fraction_simple_bounds(gen[t] / self.maximum_power)
  File "/HOPP/hopp/simulation/technologies/dispatch/power_storage/power_storage_dispatch.py", line 319, in maximum_power
    return self.blocks[t].maximum_power.value
  File "/opt/anaconda3/envs/hopp/lib/python3.8/site-packages/pyomo/core/base/param.py", line 212, in value
    return self()
  File "/opt/anaconda3/envs/hopp/lib/python3.8/site-packages/pyomo/core/base/param.py", line 867, in __call__
    return super(ScalarParam, self).__call__(exception=exception)
  File "/opt/anaconda3/envs/hopp/lib/python3.8/site-packages/pyomo/core/base/param.py", line 199, in __call__
    raise ValueError(
ValueError: Error evaluating Param value (heuristic_battery[0].maximum_power):
        The Param value is currently set to an invalid value.  This is
        typically from a scalar Param or mutable Indexed Param without`

PR #208, opened 2023-08-25
Setter bug

jaredthomas68:
Trying to get examples/optimization/hybrid_sizing_problem.py running with the recent changes to HOPP. Ran into issue #207 . When that issue is resolved we may be able to complete with pull request to get the noted example running.

### DIFF
--- a/hopp/simulation/technologies/dispatch/dispatch.py
+++ b/hopp/simulation/technologies/dispatch/dispatch.py
@@ -6,7 +6,7 @@ try:
 except AttributeError:
     u.load_definitions_from_strings(['USD = [currency]', 'lifecycle = [energy] / [energy]'])
 
-class Dispatch:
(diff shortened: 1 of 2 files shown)


hopp/simulation/technologies/dispatch/dispatch.py
```diff
@@ -6,7 +6,7 @@
 except AttributeError:
     u.load_definitions_from_strings(['USD = [currency]', 'lifecycle = [energy] / [energy]'])
 
-class Dispatch:
+class Dispatch(object):
```
camirmas, line 9: Python 3 classes implicitly inherit from `object` so we shouldn't need it here

camirmas (approved):

camirmas:
@jaredthomas68 is this meant to still be a draft?

PR #210, opened 2023-08-28
Small updates to get things working with Green Steel

ereznicek:
All edits are to hopp/to_organize/ files so this should not affect anyone else's work


jaredthomas68 (approved):
There are quite a few bits of code being changed here that are project specific. We should work on making the HOPP elements of the green steel project more general. However, for now I think we will stick with this as is.

Issue #212, opened 2023-08-30
Refactor PVPlant


camirmas:
- [ ] Convert to `attrs`
- [ ] Update docstrings/documentation
- [ ] Add baseline unit tests

PR #217, opened 2023-08-30
Refactor/wind plant

### LINKED ISSUE #211: Refactor WindPlant
- [x] Convert to `attrs`
- [ ] Update docstrings/documentation
- [ ] Add baseline unit tests
- [ ] Update YAML definition, `HoppInterface`, and `Hopp` simulation classes

bayc:
# Add HOPP Interface and Refactor WindPlant

This PR adds a new way to interact with HOPP, `HoppInterface`, which supports a yaml input file definition of the scenario you want to run in HOPP. This update also refactors the `WindPlant` class to enable modularity to allow the user to switch between two different wind plant technology models: PySAM and FLORIS.

Things left to do:

- [ ] Update FLORIS and PySAM example to run same wind farm
- [ ] Check that the outputs from using FLORIS and PySAM technology models are approximately the same
- [ ] Review unit tests for any gaps and add missing test functionality
- [ ] Confirm changes to `SiteInfo` are necessary or if left over from a previous version

## Related issue
resolves #211  

`wind_source.py` as well as new files to add the HOPP interface and input file processing.

I've updated the `simulate_hyrbid.py` example to use `HoppInterface` and have included the yaml input files. To test switching between PySAM and FLORIS, the user would update the `simulation_model` and `simulation_input_file` parameters under the wind technology definition in the yaml input file. For example:

For FLORIS:
```
  wind:
    simulation_model: floris
    simulation_input_file: "examples/inputs/floris_input.yaml"
```

For PySAM:
```
  wind:
    simulation_model: pysam
    simulation_input_file: "examples/inputs/pysam_simulation_input.yaml"
```

None


hopp/simulation/technologies/sites/site_info.py
```diff
@@ -138,7 +138,7 @@ def __attrs_post_init__(self):
         self.elec_prices = ElectricityPrices(data['lat'], data['lon'], data['year'], filepath=self.grid_resource_file)
         self.n_periods_per_day = self.n_timesteps // 365  # TODO: Does not handle leap years well
         self.interval = int((60*24)/self.n_periods_per_day)
-        self.urdb_label = data['urdb_label'] if 'urdb_label' in data.keys() else None
+        self.urdb_label = self.data['urdb_label'] if 'urdb_label' in self.data.keys() else None
```
camirmas, line 141: we shouldn't need this change due to L105. Though if we want we can simply have everything use `self.data` instead of local variable `data = self.data`

hopp/simulation/technologies/sites/site_info.py
```diff
@@ -198,13 +198,6 @@ def plot(self,
         axes.set_aspect('equal')
         axes.set(xlim=(min_plot_bound[0], max_plot_bound[0]), ylim=(min_plot_bound[1], max_plot_bound[1]))
         plot_shape(figure, axes, self.polygon, '--', color=border_color, alpha=alpha, linewidth=linewidth / 2)
-        if isinstance(self.polygon, Polygon):
```
camirmas, line 201: This is the issue I was describing yesterday, with the `shapely` version clash. We should be able to keep this code, just uninstall `shapely` from conda and make sure it's installed via `requirements.txt`

hopp/simulation/technologies/wind/pysam_wind.py
```diff
@@ -0,0 +1,65 @@
...
+@define
+class PySAMWind(BaseClass):
+    config_dict: dict = field(converter=dict)
+    site: SiteInfo = field()
+    timestep: tuple = field(default=(), converter=tuple)
+
+    system_model: Windpower = field(init=False)
+
+    def __attrs_post_init__(self):
+        input_file_path = resource_file_converter(self.config_dict["simulation_input_file"])
```
camirmas, line 20: I'm seeing this yaml file loading code used a few times (and we'll definitely be doing this for other classes), might be worth moving into its own utility function in `hopp.utilities.utilities`

hopp/simulation/technologies/wind/pysam_wind.py
```diff
@@ -0,0 +1,65 @@
...
+        else:
+            raise ValueError("Supported import filetype is YAML")
+
+        self.system_model = Windpower.new()
+        self.system_model.assign(input_dict)
+
+        self.wind_turbine_rotor_diameter = input_dict['Turbine']['wind_turbine_rotor_diameter']
+        self.wind_farm_xCoordinates = input_dict['Farm']['wind_farm_xCoordinates']
+        self.wind_farm_yCoordinates = input_dict['Farm']['wind_farm_yCoordinates']
+        self.nTurbs = len(self.wind_farm_xCoordinates)
```
camirmas, line 33: prefer `n_turbs` if possible

camirmas (commented):
looking solid so far!

hopp/simulation/hopp.py
```diff
@@ -15,7 +15,7 @@
 class Hopp(BaseClass):
     name: str = field(converter=str)
     config: dict = field(converter=dict)
-    site: SiteInfo = field()
+    site: SiteInfo = field(converter=SiteInfo.from_dict)
```
camirmas, line 18: I like this! One similar issue I ran into for config data in `PVPlant` is that the type linter gives errors when you want to supply a `dict` to be converted because it doesn't match the type signature. A quick fix is to change the type signature to `Union[dict, SiteInfo]`

hopp/simulation/technologies/sites/site_info.py
dguittet: What's happening with this? Are there problems plotting shapely objects?

camirmas: The install instructions for HOPP have this line:

`conda install -c conda-forge shapely==1.7.1 -y`

This version of shapely leads to errors with the current code. We also define `shapely` in `requirements.txt` with a more recent version, but the conda install takes precedent

hopp/simulation/technologies/wind/pysam_financial.py
```diff
@@ -0,0 +1,26 @@
...
+
+import PySAM.Singleowner as Singleowner
+
+from hopp.simulation.base import BaseClass
+from hopp.type_dec import resource_file_converter
+from hopp.utilities.utilities import load_yaml
+
+
+@define
+class PySAMFinancial(BaseClass):
```
camirmas, line 12: is this class specific to wind or can we have it live elsewhere?

hopp/simulation/technologies/wind_source.py
```diff
@@ -1,86 +1,96 @@
...
-                 farm_config: dict,
-                 rating_range_kw: tuple = (1000, 3000),
-                 ):
+@define
+class WindPlant(PowerSource, BaseClass):
+    site: SiteInfo = field()
+    farm_config: dict = field(converter=dict)
+    rating_range_kw: tuple = field(default=(1000, 3000))
+
+    _system_model: Union[PySAMWind, Floris] = field(init=False)
```
camirmas, line 34: I've been thinking a bit about the use of leading underscores on these attributes. A couple notes on this:
- `attrs` does not respect them internally: https://www.attrs.org/en/stable/init.html#private-attributes-and-aliases
- They do not really follow the "private" convention anyway, as they are used by other classes within HOPP

I'm experimenting with having these attributes not use underscores on #212

hopp/simulation/technologies/wind_source.py
```diff
@@ -1,86 +1,96 @@
...
+    }
+}
 
-    def __init__(self,
-                 site: SiteInfo,
-                 farm_config: dict,
-                 rating_range_kw: tuple = (1000, 3000),
-                 ):
+@define
+class WindPlant(PowerSource, BaseClass):
```
camirmas, line 29: I think we can have `PowerSource` inherit from `BaseClass` so that we don't have to do multiple inheritance for each refactored class

hopp/simulation/hopp.py
```diff
@@ -15,7 +15,7 @@
...
     technologies: dict = field(converter=dict)
 
     def __attrs_post_init__(self) -> None:
```
dguittet, line 21: I know this file isn't part of the pull request, but the HybridSimulation has other input configs that aren't being passed in via lines 24-28 (dispatch_options, cost_info, simulation_options)

hopp/simulation/technologies/sites/site_info.py
dguittet: If those errors are fixed, then could this plotting be added back?

examples/inputs/pysam_simulation_input.yaml
```diff
@@ -0,0 +1,495 @@
+Resource:
```
dguittet, line 1: I feel like this file's name should mention wind

hopp/simulation/technologies/wind/pysam_financial.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def __attrs_post_init__(self):
+        input_file_path = resource_file_converter(self.config_dict["financial_input_file"])
+        filetype = input_file_path.suffix.strip(".")
+        if filetype.lower() in ("yml", "yaml"):
+            input_dict = load_yaml(input_file_path)
+        else:
+            raise ValueError("Supported import filetype is YAML")
+            
+        self.financial_model = Singleowner.new()
```
dguittet, line 25: So originally when PySAM.Windpower is paired with PySAM.Singleowner, the Singleowner is created with `from_existing` instead of `new` so that the underlying data is shared between the Windpower and the Singleowner. 

https://github.com/NREL/HOPP/blob/fa6b5ab1979c28d50cb4e1b098997d563c0b25c2/hybrid/wind_source.py#L49C16-L49C16

Without that data sharing, such as for the CustomFinancialModel, there will need to be explicit functions for copying variables over: `set_financial_inputs`

https://github.com/NREL/HOPP/blob/fa6b5ab1979c28d50cb4e1b098997d563c0b25c2/hybrid/power_source.py#L59

dguittet (commented):
I would encourage writing a version of the HybridSimulation tests that use only wind such as `test_hybrid_wind_only` using the hopp interface to make sure the results are the same

hopp/simulation/technologies/wind/pysam_wind.py
dguittet: agreed

hopp/simulation/technologies/wind/pysam_wind.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        if set_value:
+            self.__setattr__(name, set_value)
+        else:
+            return self.__getattribute__(name)
+
+    def execute(self, project_life):
+        self.system_model.execute(project_life)
+
+    @property
```
dguittet, line 51: Instead of explicitly making an interface for these properties (which are only a subset of all the available ones that a user may be interested in), is it possible to automatically create them, or just pass the accessor directly to the system model. Not sure if it's possible but if `pysam_wind.annual_energy` could just return `wind_source.annual_energy`?

On a high level, I'm not sure what this new layer pysam_wind is supposed to achieve besides having the attr_post_init function that allows recursive class instantiation. If that's the point, then I think trying to make it as minimal as possible would be good, so passing all `set` and `get` to the underlying model without intermediate functions and intermediate data. But again, I'm not sure what the intention is here

hopp/simulation/technologies/wind_source.py
bayc: I agree, I think we can remove them.

hopp/simulation/technologies/wind/pysam_wind.py
bayc: Yes, barring any issues with internal names to PySAM.

hopp/simulation/technologies/wind/pysam_wind.py
bayc: Sounds good.

hopp/simulation/technologies/wind/pysam_financial.py
bayc: Need to check the input file to see if there are wind specific inputs.

hopp/simulation/hopp.py
bayc: Good call, I will add the ability to pass those inputs as well.

hopp/simulation/technologies/sites/site_info.py
bayc: I will look at adding these lines back in.

hopp/simulation/technologies/wind/pysam_financial.py
bayc: I think I need to discuss more to understand what we want to accomplish here.

hopp/simulation/technologies/wind_source.py
bayc: Agreed!

hopp/simulation/technologies/wind/pysam_financial.py
dguittet: PySAM singleowner is be technology agnostic

examples/simulate_hybrid.py
```diff
@@ -1,64 +1,11 @@
...
-from pathlib import Path
+from hopp.tools.hopp_interface import HoppInterface
 
-from hopp.simulation.technologies.sites import SiteInfo, flatirons_site
-from hopp.simulation.hybrid_simulation import HybridSimulation
-from hopp.utilities.log import hybrid_logger as logger
-from hopp.utilities.keys import set_nrel_key_dot_env
 
-examples_dir = Path(__file__).parent.absolute()
+hi = HoppInterface("inputs/wind_solar.yaml")
```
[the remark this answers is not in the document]
bayc: I will get this functionality added back in!

hopp/simulation/technologies/wind/pysam_wind.py
bayc: I look at having power source inherit from base class and see if that allows these extra properties to be removed.

hopp/simulation/technologies/wind/pysam_wind.py
bayc: Look at extra attributes on pysam_wind and see if they are still needed after updating inheritance to powersource. If so, then figure out disconnect from system_model to _system_model.

PR #218, opened 2023-08-31
Changes to use the TAP API as an alternative option to WTK

jruzekowicz:


camirmas:
Closing in favor of updated PR #237

PR #219, opened 2023-09-01
Electrolyzer degradation impacts - added features and fixes

elenya-grant:
PEM_H2_LT_electrolyzer_Clusters.py:
- Added: correct reduction in hydrogen due to degradation in off-grid cases
- Added: lifetime performance specs (capacity factor, efficiency, etc)
- Fixed: fatigue degradation bug if voltage is zero for a whole week
Future Development Plans:
- To-do: (may be fixed soon) update lifetime performance specs for grid-connected case
- To-do: update lifetime performance specs to be per year, rather than a life average
- To-do: (non-urgent) finish removing unused functions
- To-do: (non-urgent) change inputs to .yaml file and remove some hard-coded values (turn-down ratio, start-up delay, water usage, BOP energy usage, etc)

*Note that some of the functions in this file are unused or leftover.
Placeholder functions: create_system_for_target_eff(), new_calc_stack_replacement_info(), call_degradation_calculations(), check_aliveness(), make_stack_replacement_schedule(), self.estimate_life()

run_h2_PEM.py:
- Changed: run_h2_PEM() function to be compatible with new outputs from PEM_H2_LT_electrolyzer_Clusters.py
- Changed: H2_Results outputs to be reflective of lifetime performance rather than simulation performance
- Note: I embedded a dictionary in H2_Results called "new_H2_Results" which would be beneficial to switch to eventually
- Added: clean_up_final_outputs() function so that h2_ts and h2_tot (which are generally very large data frames) are reduced to some key output metrics
- removed unused imports, moved imports to functions that use them (I can undo this)

Note: unsure what the timeline is on integrating the electrolyzer model from the electrolyzer repository, but it'd be great to understand what inputs, outputs, tools, and capabilities are needed and used from the electrolyzer models in hopp and the electrolyzer repo (so that we can move towards a singular electrolyzer model)


camirmas:
If we intend to replace this with [electrolyzer](https://github.com/NREL/electrolyzer), then I'm fine merging this without documentation/testing. But like you said, we'll want to make sure we know what features and tools are needed to make that happen

camirmas (commented):
This is my first time looking through the electrolyzer code in HOPP. I'm noticing that there are lot of shared implementation details with NREL/electrolyzer, but that these implementations have a lot of duplicated code and hardcoded values. I'd like to make a push toward bringing NREL/electrolyzer in as soon as possible, and basing new development around that so that we don't diverge further. I have bandwidth to help with this, but need some help navigating the current state of things.

Let's meet up and talk about some followup stuff:
- add type annotations to classes/functions
- update docstrings ([Google format](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#google-vs-numpy))
- identify divergences between NREL/electrolyzer and this model
- NREL/electrolyzer integration steps

camirmas:
I also just noticed that there are some duplicated files in the `Electrolyzer_Models` dir, can we remove them or do they contain anything that isn't in the `electrolysis` dir?

camirmas:
Merging so we can get our examples finished. Can follow up if more changes are needed.

PR #232, opened 2023-09-27
Fix min_intrarow_spacing

dguittet:
In line 155, the `interrow_spacing` is calculated as `intrarow_spacing * grid_aspect`. So when `grid_aspect` < 1 then the interrow spacing can be less than the intrarow. The intrarow spacing is bounded by the min and max calcs in lines 178-179. The `min_intrarow_spacing` should depend on `min_spacing / grid_aspect` instead of `min_spacing * grid_aspect` so that when `interrow_spacing = intrarow_spacing * grid_aspect` happens, the min_spacing is respected.

Thanks to Maël for finding this issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ 3.7, 3.8 ]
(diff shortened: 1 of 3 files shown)


dguittet:
@camirmas Could I bug you for a rubber stamp approval?

PR #233, opened 2023-09-28
HOPP Jupyter Examples

kbrunik:
Added Jupyter Notebook examples to showcase HOPP functionality:

1. Basic Wind-Solar example with no API call
2. Basic Wind-Solar example using API call for resource data
3. Basic Wind-Solar-Battery example using default battery optimization (Profit-Maximization)
4. Basic Wind-Solar-Battery example using Load-following battery optimization
5. Use FLORIS wake modeling for wind technology
6. Basic Wave-Wind example


examples/workshop/04-load-following-battery.ipynb
```diff
@@ -0,0 +1,347 @@
+{
```
camirmas, line 1: For battery, we can specify init/min/max SOC from the top level config:

```python
technologies = {
    'pv': {
        'system_capacity_kw': solar_size_mw * 1000,
    },
    'wind': {
        'num_turbines': 10,
        'turbine_rating_kw': int(wind_size_mw * 1000 / 10)
    },
    'battery': {
        'system_capacity_kwh': battery_capacity_mw * 1000,
        'system_capacity_kw': battery_capacity_mw * 4 * 1000,
        'minimum_SOC': 20.0,
        'maximum_SOC': 100.0,
        'initial_SOC': 90.0
    },
    'grid': {
        'interconnect_kw': interconnection_size_mw * 1000
    }
}
```

Then we can remove the "Adjust Battery Parameters" section. The less that users have to manually dig into the PySAM system model, the better.

camirmas (approved):
Looks good to me, with one small change on example 4.

A couple of todos moving forward:
- After merging the refactored classes, revisit these to make sure they still work
- After completing the refactor milestone, add sections to these examples showing how to run HOPP from the YAML interface.

examples/workshop/04-load-following-battery.ipynb
kbrunik: Updated in 27f6b90b6dadbfb5911fb1788837d614e53227bf

examples/workshop/03-wind-solar-battery.ipynb
```diff
@@ -0,0 +1,325 @@
...
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from hopp import ROOT_DIR\n",
+    "from hopp.simulation.technologies.sites import SiteInfo, flatirons_site\n",
+    "\n",
+    "DEFAULT_SOLAR_RESOURCE_FILE = ROOT_DIR.parent / \"resource_files\" / \"solar\" / \"35.2018863_-101.945027_psmv3_60_2012.csv\"\n",
+    "DEFAULT_WIND_RESOURCE_FILE = ROOT_DIR.parent / \"resource_files\" / \"wind\" / \"35.2018863_-101.945027_windtoolkit_2012_60min_80m_100m.srw\"\n",
+    "DEFAULT_PRICE_FILE = ROOT_DIR.parent / \"resource_files\" / \"grid\" / \"pricing-data-2015-IronMtn-002_factors.csv\"\n",
```
dguittet, line 61: same as previous issue

examples/workshop/01-wind-solar.ipynb
```diff
@@ -0,0 +1,226 @@
...
+    "site = SiteInfo(\n",
+    "        flatirons_site,\n",
+    "        solar_resource_file=DEFAULT_SOLAR_RESOURCE_FILE,\n",
+    "        wind_resource_file=DEFAULT_WIND_RESOURCE_FILE,\n",
+    "        solar=True,\n",
+    "        wind=True,\n",
+    "        wave=False\n",
+    "    )"
+   ]
+  },
```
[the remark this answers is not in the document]
camirmas: That's a good point. My initial take is that we don't want these accessible via an official HOPP package install, as it would require us to package up data files, notebooks, and dependencies like Jupyter. However, once we publish a new version of HOPP, we can host these notebooks with something like Google Colab, which would let folks run and edit them as they explore the functionality. Open to suggestions here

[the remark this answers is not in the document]
dguittet: I think currently the whole examples folder is packaged up. I agree that they shouldn't be part of the package, so should be removed from `setup.py`. In which case the examples can only be run if you download the source code, and then the files should be available. Hosting these on Colab sounds great.

PR #236, opened 2023-10-03
Refactor CSP classes

camirmas:
# Refactor CSP classes

This PR refactors the CSP classes and their interfaces, adds new tests, and updates their documentation.

This will remain a draft until #235 is merged, as it builds off of those changes.

## Related issue

resolves #223 
resolves #222 

## Impacted areas of the software

- `csp_source.py`
- `trough_source.py`
- `tower_source.py`

## Test results, if applicable
None

PR #237, opened 2023-10-05
Tap integration

### LINKED ISSUE #228: Complete TAP integration
This is already in progress, we just need to fix the merge conflicts and update the resource tests

camirmas:
# TAP integration

This PR adds [TAP](https://www.nrel.gov/wind/tools-assessing-performance.html) as an alternative wind resource option to WTK. It returns files in the same format as WTK, so it fits into the existing resource workflow seamlessly.

This will remain a draft until #236 is merged, as it builds off of the major refactor changes.

## Related issue

resolves #228 

- `wind_resource.py`
- `solar_resource.py`

As part of this implementation, I've also added a new init option, `use_api`, which provides an override for users who want to force an API call even if a file already exists for a given resource.

[TAP API Documentation](https://dw-tap.nrel.gov/)

I've also removed some redundant resource files. We've consolidated resource files into `hopp/resource_files` as part of the refactor process.

Also removes some redundant resource files.

None


camirmas:
This will be merged via `refactor-sim-2` branch

PR #245, opened 2023-10-27
PV Plant Updates

kbrunik:
- Modified pv `capacity_factors` calculation in `HybridSimulation` to ac capacity factor by multiplying system capacity by inverter loading ratio to get kWh (ac) / kW (ac)
- Added getter/setter for dc_ac_ratio 
- Modified `PVPlant` `capacity_factor` to ac capacity factor calculation compatible with PySAM v3.0.0
- Added pv `plant_size` property
- Added pv `plant_mass` property
- Updated tests


kbrunik:
Opened a new PR #246 to develop branch

PR #246, opened 2023-10-27
Solar plant updates

kbrunik:
- Modified pv `capacity_factors` calculation in `HybridSimulation` to ac capacity factor by multiplying system capacity by inverter loading ratio to get kWh (ac) / kW (ac)
- Added getter/setter for `dc_ac_ratio`
- Modified `PVPlant` `capacity_factor` to ac capacity factor calculation compatible with PySAM v3.0.0
- Added pv `plant_area` property
- Added pv `plant_mass` property
- Updated tests

### DIFF
--- a/hopp/simulation/technologies/pv/pv_plant.py
+++ b/hopp/simulation/technologies/pv/pv_plant.py
@@ -35,7 +35,7 @@ class PVConfig(BaseClass):
 
         dc_degradation: Annual DC degradation for lifetime simulations [%/year]
         approx_nominal_efficiency: approx nominal efficiency depends on module type (standard crystalline silicon 19%, premium 21%, thin film 18%) [decimal]
-        modudule_unit_mass: Mass of the individual module unit (default to 11). [kg/m2]
+        modudule_unit_mass: Mass of the individual module unit (default to 11.092). [kg/m2]
     """
     system_capacity_kw: float = field(validator=gt_zero)
 
@@ -45,7 +45,7 @@ class PVConfig(BaseClass):
     fin_model: Optional[Union[str, dict, FinancialModelType]] = field(default=None)
     dc_degradation: Optional[List[float]] = field(default=None)
     approx_nominal_efficiency: Optional[float] = field(default=0.19)
-    module_unit_mass: Optional[float] = field(default=11)
+    module_unit_mass: Optional[float] = field(default=11.092)
 
 
 @define
@@ -110,7 +110,7 @@ class PVPlant(PowerSource):
         if self.config.module_unit_mass is not None:
             self.module_unit_mass = self.config.module_unit_mass
         else:
-            self.module_unit_mass = 11
+            self.module_unit_mass = 11.092
 
         if layout_model is not None:
             self.layout = layout_model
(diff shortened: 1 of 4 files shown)


hopp/simulation/technologies/pv/pv_plant.py
```diff
@@ -133,4 +147,60 @@ def dc_degradation(self) -> float:
...
+                raise ValueError("Module type not recognized")
+        else:
+            raise NotImplementedError("Module type not set")
+
+    @property
+    def plant_area(self):
+        """Estimated Total Module Area [m2]"""
+        if self.approx_nominal_efficiency == 0:
+            raise ValueError("approx_nominal_efficiency cannot be zero.")
+        return self._system_model.SystemDesign.system_capacity / self.approx_nominal_efficiency
```
dguittet, line 188: I don't understand why the area is the system capacity divided by an efficiency? Shouldn't it be a kw per m2 value?

kbrunik: I agree the units don't make sense, which is what I had messaged you about via Teams. This is based on an estimate in the SAM GUI for pvwatts8, we talked about getting a more specific area calculation for each module type in the future but thought this would be alright in the mean time.

dguittet: Could you use 11.092 kg/m2 as the default value, taken from the SAM Transient Thermal Model? There's even a reference if you need to explain it: Prilliman, M.; Stein, J. S.; Riley, D.;  Tamizhmani, G. (2020). [Transient Weighted Moving-Average Model of Photovoltaic Module Back-Surface Temperature](https://ieeexplore.ieee.org/document/9095219). IEEE Journal of Photovoltaics. 8 pp. Volume 10 No. 4.

kbrunik: I updated the mass calculation in https://github.com/NREL/HOPP/pull/246/commits/66d87e6f01722ac954a62d8d397d912f12532bea to use 11.092kg/m2.

Are you interested in me updating the area calculation to a different method?

dguittet: Oh yes, area! There's a file `pv_module.py` in HOPP that contains the PVWatts default module that has `module_width` and `module_height` and `module_power` so you could use that to set up the multiplier

kbrunik: Updated to use `get_module_attribs` to get PVWatts `module_area` and `module_power` in https://github.com/NREL/HOPP/pull/246/commits/721f11b65029e698acc182c4d20ffc7ec9d5295e

dguittet: thanks!

hopp/simulation/technologies/pv/pv_plant.py
```diff
@@ -1,4 +1,5 @@
 from typing import List, Sequence, Optional, Union
+from xml.sax.handler import property_declaration_handler
```
dguittet, line 2: is this used?

dguittet (approved):

PR #250, opened 2023-11-03
Update PySAM to 4.2.0

### LINKED ISSUE #248: Update PySAM to 4.2.0
Update [PySAM](https://nrel-pysam.readthedocs.io/en/main/) to the latest version. There has already been some work developing this in https://github.com/NREL/HOPP/tree/pysam_update_capacity. This should be considered a high priority, as it will enable more development on a more configurable integration with SAM.

camirmas:
resolves #248 

This builds off of work in the https://github.com/NREL/HOPP/tree/pysam_update_capacity branch, but instead updates to 4.2.0 instead of 4.1.0, and is on the new version of HOPP.  

The version change has led to some rather large differences in outputs, though I'm not super clear on what they mean.

### DIFF
--- a/hopp/simulation/hybrid_simulation.py
+++ b/hopp/simulation/hybrid_simulation.py
@@ -896,14 +896,14 @@ class HybridSimulation(BaseClass):
         return self._aggregate_financial_output("capacity_payment", 1)
 
     @property
-    def energy_purchases_values(self) -> HybridSimulationOutput:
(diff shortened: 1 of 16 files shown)


dguittet (approved):
LGTM thanks for updating all the tests!

Are there things from pysam_update_capacity that aren't included here?

camirmas:
@dguittet I cross-checked against that branch as I was doing this, but I'll do one more run-through to make sure I got everything

PR #262, opened 2023-12-27
Load following heuristic

camirmas:
# Load Following Heuristic

This PR reworks #135 for compatibility with HOPP v2.0+

I've removed the example for now, planning to make a followup to rework it in the style of our current examples (see #263).

Optimal dispatch


hopp/simulation/technologies/dispatch/power_storage/simple_battery_dispatch_heuristic.py
```diff
@@ -73,22 +73,22 @@ def _set_power_fraction_limits(self, gen: list, grid_limit: list):
     @staticmethod
     def enforce_power_fraction_simple_bounds(power_fraction) -> float:
         """ Enforces simple bounds (0,1) for battery power fractions."""
```
kbrunik, line 75: Maybe should change the docstring to 0.9 instead of 1 for the bounds. I think ideally bounds would be an assignable variable.

hopp/simulation/technologies/dispatch/power_storage/simple_battery_dispatch_heuristic.py
```diff
@@ -73,22 +73,22 @@ def _set_power_fraction_limits(self, gen: list, grid_limit: list):
...
+            power_fraction = 0
         return power_fraction
 
     def update_soc(self, power_fraction, soc0) -> float:
-        if power_fraction > 0.0:
+        if power_fraction > 0.1:
             discharge_power = power_fraction * self.maximum_power
             soc = soc0 - self.time_duration[0] * (1/(self.discharge_efficiency/100.) * discharge_power) / self.capacity
-        elif power_fraction < 0.0:
+        elif power_fraction < 0.9:
```
kbrunik, line 86: Doesn't this `if` `elif` statement mean that it will charge only if the `power_fraction` is less than 0.1, is that correct?

hopp/simulation/technologies/dispatch/power_storage/simple_battery_dispatch_heuristic.py
```diff
@@ -73,22 +73,22 @@ def _set_power_fraction_limits(self, gen: list, grid_limit: list):
...
-            power_fraction = 1.0
-        elif power_fraction < 0.0:
-            power_fraction = 0.0
+        if power_fraction > 0.9:
+            power_fraction = 0.9
+        elif power_fraction < 0:
+            power_fraction = 0
         return power_fraction
 
     def update_soc(self, power_fraction, soc0) -> float:
```
kbrunik, line 82: Add docstring with charging/discharging limits.

hopp/simulation/technologies/grid.py
```diff
@@ -125,8 +130,66 @@ def simulate_grid_connection(
...
             self.schedule_curtailed_percentage = sum(self.schedule_curtailed)/sum(lifetime_schedule)
+
+            # NOTE: This is currently only happening for load following, would be good to make it more general
+            #           i.e. so that this analysis can be used when load following isn't being used (without storage)
+            #           for comparison 
+            N_hybrid = len(self.generation_profile)
+
+            final_power_production = total_gen
+            schedule = [x for x in lifetime_schedule]
+            print(len(final_power_production), len(schedule))
```
kbrunik, line 141: Do we want to leave all of these print statements in?

hopp/simulation/technologies/grid.py
```diff
@@ -125,8 +130,66 @@ def simulate_grid_connection(
...
+
+            load_met = len([i for i in hybrid_power if i  >= 0])
+            self.time_load_met = 100 * load_met/N_hybrid
+
+            final_power_array = np.array(final_power_production)
+            power_met = np.where(final_power_array > schedule, schedule, final_power_array)
+            self.capacity_factor_load = np.sum(power_met) / np.sum(schedule) * 100
+
+            print('Percent of time firm power requirement is met: ', np.round(self.time_load_met,2))
+            print('Percent total firm power requirement is satisfied: ', np.round(self.capacity_factor_load,2))
```
kbrunik, line 152: print statements

hopp/simulation/technologies/grid.py
```diff
@@ -125,8 +130,66 @@ def simulate_grid_connection(
...
+
+                group_starts = np.where(edge_mask == 1)[0]
+                group_stops = np.where(edge_mask == -1)[0]
+
+                # Find groups and drop groups that are too small
+                groups = [group for group in zip(group_starts,group_stops) if ((group[1]-group[0]) >= min_regulation_hours)]
+                group_lengths = [len(final_power_production[group[0]:group[1]]) for group in groups]
+                self.total_number_hours = sum(group_lengths)
+
+                print('Total number of hours available for ERS: ', np.round(self.total_number_hours,2))
```
kbrunik, line 190: print statement

hopp/simulation/technologies/grid.py
camirmas: Really these should be part of the logger, we have this problem in a number of places in the codebase. I'm not sure the logger is working as intended right now, but I'll add them as logs anyway

hopp/simulation/technologies/dispatch/power_storage/simple_battery_dispatch_heuristic.py
camirmas: I agree. I think this is part of a bigger problem where need more configurability at the dispatch level. A quick fix would be to add power fraction args to `HybridDispatchOptions`, but it might be worth a more serious look

camirmas: I agree. I think this is part of a bigger issue where we need more configurability for the dispatch, with varying config options for different strategies. A quick fix would be to add power fraction options to `HybridDispatchOptions`, but I think it deserves a bit more work than that

hopp/simulation/technologies/dispatch/power_storage/simple_battery_dispatch_heuristic.py
camirmas: Yeah this is is kinda misleading. It's equivalent to saying `elif power_fraction < 0.1`, so maybe that's a clearer way to express this

kbrunik (approved):

PR #265, opened 2024-01-17
Profast

jaredthomas68:
# Include ProFAST `.pyc` files

This includes profast in the HOPP repository and includes all necessary files, including the unique license file for ProFAST. Note that the profast `.pyc` files included here are only for python version 3.8.

- Module ProFAST has been added
- LICENSE file has been modified
- Two tests have been added
- setup.py file has been modified to include the ProFAST module
- 

Capitalization is as indicated by ProFAST dev team, but we may be able to modify it after checking with them.


kbrunik (approved):
Will need to fix build since it's failing, will address in subsequent PR.

PR #266, opened 2024-01-17
ProFAST Example

kbrunik:
- Added basic example detail how to use ProFAST within HOPP
- Fixed required ProFAST imports

ProFAST is only compatible with python=3.8. Builds 3.9 and 3.10 in CI fail.


camirmas (approved):
Looks good, merge when 3.8 build passes

PR #271, opened 2024-02-20
ARIES Capabilities

kbrunik:
Added:

- Ability to get `turb_powers`, which is the power from the individual turbines.
- Ability to get `turb_velocities`, which is the effective rotor velocities, which are an average of the incoming velocity points on the rotor.
- Added `output_level` into `hybrid_system.py` to access additional pieces of information required for ARIES work.

Modified:

- Update 18MW turbine with power-curve scaled [IEA 15MW turbine](https://github.com/IEAWindTask37/IEA-15-240-RWT).
- Added method to remove `nan` from ORBIT layout when it's input into FLORIS layout.
- Updated values in `test_eco.py` from changes to turbine power curve and fix around removing `nan` from FLORIS.
- Removed hardcoded losses applied to `power_farm` and added it as an input `operational_losses` to `WindPlant` with the same default as hardcoded value.
- Applied `operational_losses` to `gen` and `turb_powers`.


kbrunik:
I realized that it's likely that the individual turbine powers (`turb_powers`) from `FLORIS` will not sum to the `gen` because there's a hard coded 12.83% loss in the `execute` fuction in `FLORIS`

tests/greenheart/test_hydrogen/test_eco.py
```diff
@@ -36,13 +37,13 @@ def setUpClass(self):
         filename_eco_config = os.path.join(orbit_library_path, f"plant/eco_config.yaml")
         filename_hopp_config = os.path.join(orbit_library_path, f"plant/hopp_config.yaml")
 
-        self.lcoe, self.lcoh, _ = run_simulation(filename_hopp_config, filename_eco_config, filename_turbine_config, filename_orbit_config, filename_floris_config, verbose=False, show_plots=False, save_plots=False,  use_profast=True, incentive_option=1, plant_design_scenario=1, output_level=4)
+        self.lcoe, self.lcoh, _ = run_simulation(filename_hopp_config, filename_eco_config, filename_turbine_config, filename_orbit_config, filename_floris_config, verbose=False, show_plots=False, save_plots=False,  use_profast=True, post_processing=True, incentive_option=1, plant_design_scenario=1, output_level=4)
 
     def test_lcoh(self):
-        assert self.lcoh == approx(5.70230272215567) # TODO base this test value on something
+        assert self.lcoh == approx(7.057994298481547) # TODO base this test value on something
```
jaredthomas68, line 43: It looks like the change increased the LCOH and LCOE estimate considerably. They only change that should have had any impact was the loss percentage. I think it would be good to try setting the loss to 0.0 and just make sure the test results go back to the values before adding the default loss.

jaredthomas68 (changes requested):

tests/greenheart/test_hydrogen/test_eco.py
kbrunik: The default loss was already being applied it was just a hardcoded value. The main difference in lcoh is coming from correcting the `nan` coming from ORBIT. FLORIS was interpreting those as turbines so for the test FLORIS was giving power for 12 turbines instead of 10.

PR #274, opened 2024-02-23
Add Steel/Ammonia models to greenheart tech

camirmas:
# Steel/Ammonia

This PR takes steel and ammonia models represented in the green steel project work and updates them for use within the current greenheart framework. Both models have three functions:
- production model: for now, this is just based on capacity/capacity factor
- cost model: calculates capex, opex, installation
- finance model: runs finances with ProFAST

Each function has dataclasses associated with inputs and outputs, providing a clear interface that includes relevant defaults (though these defaults should be investigated in more detail).


greenheart/simulation/technologies/ammonia/ammonia.py
```diff
@@ -0,0 +1,694 @@
...
+            factor, and feedstock costs.
+
+    Returns:
+        AmmoniaCostModelOutputs: Object containing detailed breakdowns of calculated
+            costs, including total capital expenditure, operating costs, and credits
+            from byproducts.
+    """
+    feedstocks = config.feedstocks
+
+    model_year_CEPCI = 596.2  # TODO: what year
```
jaredthomas68, line 178: Let's get this figured out soon. Either prior to merge or make an issue to make sure this does not get forgotten.

greenheart/simulation/technologies/ammonia/ammonia.py
```diff
@@ -0,0 +1,694 @@
...
+    # ------------------------------ Sovle for breakeven price ---------------------------
+
+    sol = pf.solve_price()
+
+    summary = pf.get_summary_vals()
+
+    price_breakdown = pf.get_cost_breakdown()
+
+    price_breakdown_air_separation_by_cryogenic = price_breakdown.loc[
+        price_breakdown["Name"] == "Air Separation by Cryogenic", "NPV"
```
jaredthomas68, line 542: The price breakdown code seems more complicated than necessary. Why do we need to pull out each item individually?

greenheart/simulation/technologies/steel.py
```diff
@@ -0,0 +1,970 @@
...
+        operating costs, as well as total plant cost and other financial metrics.
+
+    Note:
+        The calculation includes various cost components such as electric arc furnace
+        (EAF) casting, shaft furnace, oxygen supply, hydrogen preheating, cooling tower,
+        and more, adjusted based on the Chemical Engineering Plant Cost Index (CEPCI).
+    """
+    feedstocks = config.feedstocks
+
+    model_year_CEPCI = 596.2
```
jaredthomas68, line 238: As in ammonia model, we should at least make an issue for setting these values/figureing out what year these are for

jaredthomas68: Better to just ask Evan the year and move on

greenheart/simulation/technologies/steel.py
```diff
@@ -0,0 +1,970 @@
...
+
+    # apply all params passed through from config
+    for param, val in config.financial_assumptions.items():
+        print(f"setting {param}: {val}")
+        pf.set_params(param, val)
+
+    analysis_start = int(list(config.grid_prices.keys())[0]) - config.install_years
+
+    # Fill these in - can have most of them as 0 also
+    gen_inflation = 0.00
```
jaredthomas68, line 546: I think gen_inflation should be an input so we can do real or nominal costs

greenheart/simulation/technologies/steel.py
```diff
@@ -0,0 +1,970 @@
...
+        + price_breakdown_iron_ore
+        + price_breakdown_hydrogen
+        + price_breakdown_natural_gas
+        + price_breakdown_electricity
+        + price_breakdown_slag
+        + price_breakdown_taxes
+        + price_breakdown_financial_equipment
+        + price_breakdown_financial_remaining
+        + price_breakdown_O2sales
+    )  # a neater way to implement is add to price_breakdowns but I am not sure if ProFAST can handle negative costs
```
jaredthomas68, line 929: Again, profast provides a price breakdown, so I'm not sure why all this code pulling the individual parts.

greenheart/simulation/technologies/ammonia/ammonia.py
```diff
@@ -0,0 +1,694 @@
...
+    pf = ProFAST.ProFAST("blank")
+
+    # apply all params passed through from config
+    for param, val in config.financial_assumptions.items():
+        pf.set_params(param, val)
+
+    analysis_start = int(list(config.grid_prices.keys())[0]) - config.install_years
+
+    # Fill these in - can have most of them as 0 also
+    gen_inflation = 0.00
```
jaredthomas68, line 382: This needs to be an input so we can choose whether to do nominal or real prices

jaredthomas68 (changes requested):
Great work! I just left a few mostly minor comments.

greenheart/simulation/technologies/ammonia/ammonia.py
camirmas: I agree, I was wondering about this, I left it as it was so we could decide if we needed it for some reason

greenheart/simulation/technologies/steel.py
camirmas: I'm not really sure about this either, should we remove it?

PR #278, opened 2024-02-28
Add end-use sizing

kbrunik:
- Add ammonia plant sizing
- Add steel plant sizing
- Update ammonia test
- Update steel test


greenheart/simulation/technologies/ammonia/ammonia.py
```diff
@@ -156,6 +156,70 @@ def run_ammonia_model(
...
+        ammonia_plant_size_kgpy = (config.hydrogen_amount_kg 
+            / config.feedstocks.hydrogen_consumption 
+            * config.plant_capacity_factor
+        )
+        hydrogen_amount_kg = config.hydrogen_amount_kg
+
+    if config.ammonia_plant_size_kgpy:
+        hydrogen_amount_kg = (config.ammonia_plant_size_kgpy 
+            * config.feedstocks.hydrogen_consumption
+            / config.plant_capacity_factor
```
jaredthomas68, line 213: We should probably call this something like `input_capacity_factor_estimate` or something so no one takes it as an output.

jaredthomas68: We need to be clear about what capacity factor we are estimating here. Is this the capacity factor of the ammonia plant or the energy plant?

jaredthomas68: I think your if statements here may be flipped. If the user provides the hydrogen amount, the sizing should return the ammonia amount, and vice versa.

greenheart/simulation/technologies/steel/steel.py
```diff
@@ -191,6 +191,72 @@ class SteelCostModelOutputs(SteelCosts):
...
+            the size of the steel plant in metric tonnes per year is output.
+        hydrogen_amount_kg (float): If amount of steel production in metric tonnes per year is input, 
+            the amount of necessary hydrogen feedstock in kilograms per year is output.
+    """
+    steel_plant_size_mtpy: float
+    hydrogen_amount_kg: float
+
+def run_size_steel_plant(config: SteelSizeModelConfig) -> SteelSizeModelOutputs:
+    
+    if config.hydrogen_amount_kg:
```
jaredthomas68, line 238: Same comment as above on possibly flipped if statements

greenheart/simulation/technologies/steel/steel.py
```diff
@@ -191,6 +191,72 @@ class SteelCostModelOutputs(SteelCosts):
...
+        steel_plant_size_mtpy = (config.hydrogen_amount_kg 
+            / 1000
+            / config.feedstocks.hydrogen_consumption 
+            * config.plant_capacity_factor
```
jaredthomas68, line 242: same comment as above on capacity factor name

jaredthomas68 (changes requested):

tests/greenheart/test_steel.py
elenya-grant, line 1: can we change the function name from run_size_steel_plant_capcity to run_size_steel_plant_capacity?

jaredthomas68 (approved):

PR #279, opened 2024-02-29
Adding Green Steel compressor test to greenheart tests

genevievestarke:
Updating the compressor capex test to sync up with Green Steel models and analysis


tests/greenheart/test_hydrogen/test_h2_transport/test_h2_compressor.py
jaredthomas68, line 1: After talking with Masha it sounds like we should try to check against a published source of our model rather than the Greensteel code as their values are pretty rough estimates.

jaredthomas68 (changes requested):

bayc:
@kbrunik Per meeting, check value against HDSAM.

PR #288, opened 2024-03-13
Greenheart

jaredthomas68:
# Enable PV and battery ratings in OpenMDAO-wrapped HOPP

PV and battery ratings may now be specified as design variables in OpenMDAO-wrapped HOPP. This is done by passing in an optional list that specifies the design variables to be instantiated.

openmdao.py and tests, as well as an implementation of a reinitialize method in the hopp interface

All tests are passing

### DIFF
--- a/hopp/simulation/hopp_interface.py
+++ b/hopp/simulation/hopp_interface.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 from pathlib import Path
 from typing import Union, TYPE_CHECKING
 
-from hopp.simulation.hopp import Hopp
+from hopp.simulation.hopp import Hopp, SiteInfo
 
 # avoid potential circular dep
 if TYPE_CHECKING:
@@ -42,8 +42,8 @@ class HoppInterface:
         elif isinstance(self.configuration, dict):
             self.hopp = Hopp.from_dict(self.configuration)
 
-    def reinitialize(self):
-        pass
+    def reinitialize(self, configuration: Union[dict, str, Path]):
+        self.__init__(configuration)
 
     def simulate(self, project_life: int = 25, lifetime_sim: bool = False):
         self.hopp.simulate(project_life, lifetime_sim)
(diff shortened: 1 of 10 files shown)


hopp/simulation/hopp_interface.py
```diff
@@ -42,8 +42,8 @@ def __init__(self, configuration: Union[dict, str, Path]):
         elif isinstance(self.configuration, dict):
             self.hopp = Hopp.from_dict(self.configuration)
 
-    def reinitialize(self):
-        pass
+    def reinitialize(self, configuration: Union[dict, str, Path]):
+        self.__init__(configuration)
```
camirmas, line 46: Small nitpick here. It's not unsafe to call `__init__` manually, but it's also not conventional Python. Alternatively you could pull the init logic into `reinitialize` and have `__init__` call `reinitialize`, or could have a separate method like `reset` that is called from both `__init__` and `reinitialize`

tests/greenheart/test_openmdao.py
```diff
@@ -1,23 +1,28 @@
...
+
+import unittest
 
 solar_resource_file = Path(__file__).absolute().parent.parent.parent / "resource_files" / "solar" / "35.2018863_-101.945027_psmv3_60_2012.csv"
 wind_resource_file = Path(__file__).absolute().parent.parent.parent / "resource_files" / "wind" / "35.2018863_-101.945027_windtoolkit_2012_60min_80m_100m.srw"
 floris_input_file = Path(__file__).absolute().parent / "inputs" / "floris_input.yaml"
 hopp_config_filename = Path(__file__).absolute().parent / "inputs" / "hopp_config.yaml"
 
-class TestBoundaryDistanceComponent():
+class TestBoundaryDistanceComponent(unittest.TestCase):
```
camirmas, line 20: It would be better to have these use pytest/subtest tools and conventions to be consistent with the rest of the test suite, but this can be done later if we're short on time

camirmas (approved):

hopp/simulation/hopp_interface.py
jaredthomas68: Thanks, I have now fixed this.

Issue #284, opened 2024-03-08
GS om cost input approach needs to be consistent across technologies and work with pysam and custom financial model


bayc:
Starting with branch: https://github.com/jaredthomas68/HOPP/tree/greenheart

PR #289, opened 2024-03-15
Update custom financial model dictionary input form

bayc:
# Expand `custom_financial_model` to use a more well-defined dictionary and utilize defaults for the `default_fin_config`

This PR updates `custom_financial_model` to use a dictionary that separates the configuration dictionary by technology. It also established default values within the code so the user does not have to supply the `default_fin_config.yaml` explicitly, with the one exception being the `financial_parameters`. This simplifies setup and more easily enables a user to update specific parts of the default financial configuration for different technologies.

## Related issue
#284 

`custom_financial_model.py`
Tests

This method still requires the user to specify a dictionary to use the custom financial model. In the inputs files, where we don't want to specify the input explicitly or point to a yaml file, the user will need to supply at the bare minimum a `financial_parameters` dictionary with both `inflation_rate` and `real_discount_rate` defined, as shown below.

```
financial_parameters:
  inflation_rate: 2.5
  real_discount_rate: 6.4
```

Tests are passing.


jaredthomas68 (approved):
Looks good to me

camirmas (approved):

PR #291, opened 2024-03-19
Bug Fix: email for API calls

elenya-grant:
# Bug Fix: user email for API calls

The resource base class (hopp/simulation/technologies/resource/resource.py) had an email attribute that was hard-coded and no longer works for API calls. This email was used in the url for API calls to WIND toolkit (wind_resource.py) and NSRDB (solar_resource.py) and should be the user's email that they used to get their API key. 

Now - the function set_nrel_key_dot_env() sets both the user email and api key. The user email is now read from the same .env file that the API key is set in.

## Related issue

Unsure if an issue exists - but it was discussed.

- **hopp/utilities/keys.py**: added `NREL_API_EMAIL` to be loaded in `set_nrel_key_dot_env()`
- **hopp/utilities/keys.py**: added `get_developer_nrel_gov_email()` and `set_developer_nrel_gov_email()`
- **hopp/simulation/technologies/resource/resource.py**: removed `self.email`
- **hopp/simulation/technologies/resource/wind_resource.py**: replaced email value in url string to user email, similar format as setting api key.
- **hopp/simulation/technologies/resource/solar_resource.py**: replaced email value in url string to user email, similar format as setting api key.
- **README.md**: updated instructions for .env file to include user email.

- **Existing users will need to update their .env file.** 

### Future Development:
Add a tests to check that:
1. The user has correctly set up the .env and in the correct location
2. Wind and solar resource downloads that require API calls do not throw errors and the files are downloaded properly

Tests should pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ solar and storage.
     Create a file ".env" that contains the single line:
     ```
     NREL_API_KEY=key
-    EMAIL=your.name@email.com
+    NREL_API_EMAIL=your.name@email.com
     ```
 
 8. Verify setup by running tests:
--- a/README.md
+++ b/README.md
@@ -59,12 +59,12 @@ solar and storage.
     [https://developer.nrel.gov/signup/](https://developer.nrel.gov/signup/)
     
 
-7. To set up the `NREL_API_KEY` required for resource downloads, you can create an Environment Variable called 
-   `NREL_API_KEY`. Otherwise, you can keep the key in a new file called ".env" in the root directory of this project. 
+7. To set up the `NREL_API_KEY` and `NREL_API_EMAIL` required for resource downloads, you can create Environment Variables called `NREL_API_KEY` and `NREL_API_EMAIL`. Otherwise, you can keep the key in a new file called ".env" in the root directory of this project. 
 
     Create a file ".env" that contains the single line:
-     ```
+    ```
     NREL_API_KEY=key
+    EMAIL=your.name@email.com
     ```
 
 8. Verify setup by running tests:
--- a/hopp/simulation/technologies/resource/solar_resource.py
+++ b/hopp/simulation/technologies/resource/solar_resource.py
@@ -6,7 +6,7 @@ import numpy as np
 import csv
 from PySAM.ResourceTools import SAM_CSV_to_solar_data
 
-from hopp.utilities.keys import get_developer_nrel_gov_key
+from hopp.utilities.keys import get_developer_nrel_gov_key,get_developer_nrel_gov_email
 from hopp.utilities.log import hybrid_logger as logger
 from hopp.simulation.technologies.resource.resource import Resource
 from hopp import ROOT_DIR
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -59,12 +59,12 @@ solar and storage.
...
 
-7. To set up the `NREL_API_KEY` required for resource downloads, you can create an Environment Variable called 
-   `NREL_API_KEY`. Otherwise, you can keep the key in a new file called ".env" in the root directory of this project. 
+7. To set up the `NREL_API_KEY` and `NREL_API_EMAIL` required for resource downloads, you can create Environment Variables called `NREL_API_KEY` and `NREL_API_EMAIL`. Otherwise, you can keep the key in a new file called ".env" in the root directory of this project. 
 
     Create a file ".env" that contains the single line:
-     ```
+    ```
     NREL_API_KEY=key
+    EMAIL=your.name@email.com
```
jaredthomas68, line 67: I think it will create confusion to use "EMAIL" in the file, but "NREL_API_EMAIL" for the environment variable name. I think we should use "NREL_API_EMAIL" in both places.

hopp/simulation/technologies/resource/solar_resource.py
```diff
@@ -6,7 +6,7 @@
 import csv
 from PySAM.ResourceTools import SAM_CSV_to_solar_data
 
-from hopp.utilities.keys import get_developer_nrel_gov_key
+from hopp.utilities.keys import get_developer_nrel_gov_key,get_developer_nrel_gov_email
```
jaredthomas68, line 9: need space after comma separator

hopp/simulation/technologies/resource/wind_resource.py
```diff
@@ -3,7 +3,7 @@
 from typing import Union
 from PySAM.ResourceTools import SRW_to_wind_data
 
-from hopp.utilities.keys import get_developer_nrel_gov_key
+from hopp.utilities.keys import get_developer_nrel_gov_key,get_developer_nrel_gov_email
```
jaredthomas68, line 6: space after comma

jaredthomas68 (changes requested):
A general question for everyone: can only NREL employees use the developer key and email? If others can use them, then we should change the function names I think.

elenya-grant:
> A general question for everyone: can only NREL employees use the developer key and email? If others can use them, then we should change the function names I think.

I think anyone can use them - but the key is kind of "from" NREL and then the email is the one associated with that key. I agree that we could rename them, but it is an API key to access NREL data (not an NREL employee's API key)

README.md
elenya-grant: I made these changes but I'm not seeing these updates reflected in the PR.

jaredthomas68 (approved):
I think I'm satisfied with where this is at. I would like to change the "gov" part of the email and api functions, but I think most people will be using the `set_nrel_key_dot_env` function and won't see the other function names anyway.

PR #300, opened 2024-03-29
Greensteel eco sync only run orbit post processing if we used orbit

jaredthomas68:
# Greensteel eco sync only run orbit post processing if we used orbit

Greensteel eco sync only run orbit post processing if we used orbit

greenheart/tools/utilities.py -> post_process_simulation()


greenheart/tools/eco/utilities.py
```diff
@@ -1237,80 +1237,81 @@ def post_process_simulation(
...
+            )
+
+        # save ORBIT cost information
+        ob_df = pd.DataFrame(orbit_capex_breakdown, index=[0]).transpose()
+        savedir = "data/orbit_costs/"
+        if not os.path.exists(savedir):
+            os.makedirs(savedir)
+        ob_df.to_csv(
+            savedir
+            + "orbit_cost_breakdown_with_onshore_substation_lcoh_design%i_incentive%i_%sstorage.csv"
```
kbrunik, line 1308: Would be nice to add lat, lon or something to be able to save these breakdowns for different locations too.

kbrunik (approved):

kbrunik:
Just one small comment on saving the orbit files, but otherwise looks good.

greenheart/tools/eco/utilities.py
jaredthomas68: I agree, but I think that will need to be up to the user changing the output directory at this point.

PR #309, opened 2024-04-18
add comment on singlitico electrolyzer cost model

jaredthomas68:
# Add meaningful title here

Describe your feature here.


kbrunik (approved):

kbrunik:
It could be nice to include some of that information in the docstring so users can see that explicitly when using the model.

Issue #283, opened 2024-03-06
Create greenheart input/output configs


camirmas:
# greenheart input/output configs

Currently, `run_simulation` takes many arguments, including filenames, which are then loaded and used to run the simulation. This makes programmatic runs (i.e. optimization) more difficult and time consuming.

We are mixing concerns between interface and implementation. This leads to issues where there is no way to specify whether a value should be taken from the config, or calculated within the simulation. These side effects are difficult to track down and may cause misleading results for researchers.

## Proposed solution

We need to be able to run the greenheart simulation in a more flexible way. The file handling and other configuration elements should be pulled out so that we may split the configuration/interface from the implementation. We have something similar in the `Hopp` -> `HybridSimulation` workflow, so we can use a similar pattern here:

- `GreenHeartConfig`: dataclass responsible for the interface to `run_simulation`. The initialization logic can live on this class, or in some function like `configure_simulation`
- `run_simulation`: will now take `GreenHeartConfig` as an argument
- `GreenHeartOutputs`: dataclass that encapsulates relevant outputs

Within run_simulation, we can then utilize the config to determine whether fields should be calculated or user-specified.

PR #312, opened 2024-04-29
Uniform output

jaredthomas68:
# Uniform flow output

This pull request fixes a bug where not all flows in the system were of uniform length. It also provides a file output with all electricity and hydrogen flows. I also took the opportunity to make a new file for testing the greenheart utilities and moved the greenheart system test out of the test_hydrogen folder and up to the top-level greenheart test directory to match the package file structure.

Another, larger change in the PR is the addition of the GreenHEARTOutput data class. Hopefully this data class will eventually take the place of the different output levels (though we may keep some of the output levels for optimization purposes).

## Related issue

This PR may close, or get close to closing, issue #283

greenheart_simulation and  greenheart testing


greenheart/simulation/greenheart_simulation.py
```diff
@@ -818,7 +887,30 @@ def simple_solver(initial_guess=0.0):
         return hopp_results, electrolyzer_physics_results, remaining_power_profile
     elif config.output_level == 7:
         return lcoe, lcoh, steel_finance, ammonia_finance
-
+    elif config.output_level == 8:
```
kbrunik, line 890: What do you think about getting rid of `output_level` completely and just returning the data class? I know that it might mess a few folks up right now but lots of other breaking changes are happening on this branch and I think using data classes will help improve the greenheart simulation generalization

greenheart/tools/eco/utilities.py
```diff
@@ -1305,30 +1307,41 @@ def save_power_series(
...
+    
     df = pd.DataFrame.from_dict(output)
 
     filepath = os.path.abspath(output_dir + "data/production/")
+
     if not os.path.exists(filepath):
         os.makedirs(filepath)
 
-    df.to_csv(os.path.join(filepath, "power_series.csv"))
+    df.to_csv(os.path.join(filepath, "energy_flows.csv"))
```
kbrunik, line 1342: I think as we keep generating more figures and data files, we should try to settle on what sort of information should be included in the file name. This format only works in the single run case which can make running sweeps difficult

tests/greenheart/test_openmdao.py
```diff
@@ -18,10 +18,10 @@
...
-greenheart_config_onshore_filename = Path(__file__).absolute().parent / "test_hydrogen" / "input_files" / "plant" / "greenheart_config_onshore.yaml"
-turbine_config_filename = Path(__file__).absolute().parent / "test_hydrogen" / "input_files" / "turbines" / "osw_18MW.yaml"
-floris_input_filename_steel_ammonia = Path(__file__).absolute().parent / "test_hydrogen" / "input_files" / "floris" / "floris_input_osw_18MW.yaml"
+hopp_config_steel_ammonia_filename = Path(__file__).absolute().parent / "input_files" / "plant" / "hopp_config.yaml"
+greenheart_config_onshore_filename = Path(__file__).absolute().parent / "input_files" / "plant" / "greenheart_config_onshore.yaml"
+turbine_config_filename = Path(__file__).absolute().parent / "input_files" / "turbines" / "osw_18MW.yaml"
+floris_input_filename_steel_ammonia = Path(__file__).absolute().parent / "input_files" / "floris" / "floris_input_osw_18MW.yaml"
 rtol = 1E-5
 
 class TestBoundaryDistanceComponent(unittest.TestCase):
```
kbrunik, line 27: Rewrite as pytests?

kbrunik (approved):
I have left a few comments for discussion, otherwise looks good.

greenheart/simulation/greenheart_simulation.py
jaredthomas68: I would like to avoid running post-processing and passing a lot of extra stuff around during optimization runs. So, for now anyway, I would like to keep the output levels. We should probably make output level 8 the de facto in our example yamls though to encourage people to use that option

greenheart/tools/eco/utilities.py
jaredthomas68: I agree, data and figure handling is a mess and should be cleaned up, unified, and generalized.

jaredthomas68: I think should be made an issue so we can address it, but not in this PR

tests/greenheart/test_openmdao.py
jaredthomas68: Right, good catch. I told you I would get this into this PR and forgot. I'll try to get to it soon.

jaredthomas68: I have adjusted the test style in `test_openmdao.py` to follow the function pytest approach

PR #313, opened 2024-04-29
GreenHEART Simulation Examples

kbrunik:
# GreenHEART Simulation Examples

Added Jupyter notebook examples to the examples folder.

1. Green steel example
2. Offshore green hydrogen example


jaredthomas68:
I think this looks pretty good. For the version of these that will help act as documentation to GreenHEART, let's remove the images from the repo so we don't clutter the git history. I think you also need to do a back-merge from the upstream branch to resolve the conflicts.

jaredthomas68 (changes requested):
This looks good, but for the version we pull into the main repo let's remove all the outputs (data, figures, etc) from the git history. Also, please back-merge to fix the conflict.

Thanks!

PR #316, opened 2024-05-14
Greenheart MPI

jaredthomas68:
# Enable OpenMDAO MPI capabilities for GreenHEART

This PR enables users to leverage the MPI capabilities of GreenHEART to use parallel finite differencing during optimization and run cases in parallel during a design of experiments.

greenheart

The necessary installation steps to run in parallel are the same as for OpenMDAO

No added or changed tests

### DIFF
--- a/greenheart/tools/optimization/gc_run_greenheart.py
+++ b/greenheart/tools/optimization/gc_run_greenheart.py
@@ -1,16 +1,21 @@
+"""
+This file is based on the WISDEM file of the same name: https://github.com/WISDEM/WISDEM
+"""
+
 import os
 import sys
 import logging
 import warnings
 
 import numpy as np
-from mpi4py import MPI
 import openmdao.api as om
 
 
 from greenheart.simulation.greenheart_simulation import GreenHeartSimulationConfig, setup_greenheart_simulation
 from greenheart.tools.optimization.gc_PoseOptimization import PoseOptimization
 from greenheart.tools.optimization.openmdao import GreenHeartComponent
+from greenheart.tools.optimization.mpi_tools import MPI, map_comm_heirarchical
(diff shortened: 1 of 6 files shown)


greenheart/tools/optimization/gc_run_greenheart.py
```diff
@@ -1,16 +1,21 @@
+"""
+This file is based on the WISDEM file of the same name: https://github.com/WISDEM/WISDEM
```
kbrunik, line 2: I don't think WISDEM has a file called gc_run_greenheart?

greenheart/tools/optimization/gc_run_greenheart.py
```diff
@@ -28,35 +33,40 @@ def run_greenheart(config:GreenHeartSimulationConfig, overridden_values=None, ru
     # Otherwise, initialize the WindPark system normally. Get the rank number for parallelization. We only print output files using the root processor.
```
kbrunik, line 33: Not sure if we want to update this comment since I'm not sure if it's the WindPark system we are parallelizing?

kbrunik (approved):
Looks like there's a little redundancy with PR #315 so let's be sure to get that in first. No major comments but would be good to address before pulling this in

greenheart/tools/optimization/gc_run_greenheart.py
jaredthomas68: Right, fixed. Thanks

greenheart/tools/optimization/gc_run_greenheart.py
jaredthomas68: Good point. I have updated the comment.

PR #318, opened 2024-05-20
make derivative specs optional for GreenHEART optimization interface

jaredthomas68:
# Make derivate specs optional so using gradient-free optimization algorithms is more straight forward

This PR makes the derivative type and form optional inputs for optimization so that running with gradient-free optimization algorithms like COBYLA and NSAGAII is more intuitive.


kbrunik (approved):
It could be good to create an issues to make an attrs class to validate the greenheart_config before running the simulation. Then it will help with documenting the different options that are being added for optimization. Right now we are too dependent on the example greenheart_config files which aren't being updated with each change

PR #321, opened 2024-05-23
Unify resource files location

jaredthomas68:
# Unify resource file location

There were two locations in HOPP where resource files were being stored: (1) `HOPP/resource_files/`, (2) `HOPP/hopp/simulation/resource_files/`. This pull request moves all necessary files from (1) to (2) and removes (1). All references to (1) have been redirected to (2).

## Related issue

While an official issue was never made, location (1) was problematic because the data in (1) could not be directly referenced when HOPP was installed on an HPC system and could likely cause problems in other environments. Moving all resource files to (2) enables storing the resource data files inside the installed hopp package and will thus be available for all installation approaches and environments.


hopp/simulation/technologies/reopt.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, lat, lon,
         # paths
         self.path_current = os.path.dirname(os.path.abspath(__file__))
         self.path_results = os.path.join(self.path_current)
-        self.path_rates = os.path.join(self.path_current, '..', 'resource_files', 'utility_rates')
+        self.path_rates = os.path.join(self.path_current, '../', 'resource_files', 'utility_rates')
```
dguittet, line 79: extra / here

dguittet (approved):
Looks great! just a small change

PR #322, opened 2024-05-24
Syncing GreenSteel and GreenHEART features

kbrunik:
# Updating GreenHEART features based on GreenSteel project code

Describe your feature here.

Several tests in the `test_greensteel.py` file are commented out. They will fail if uncommented and they are left commented out to illustrate the differences between GreenSteel and GreenHEART frameworks.


jaredthomas68 (changes requested):
This looks good to me once we get the keys changes in greenheart_simulation figured out.

examples/reference_plants/05-offshore-hydrogen-ca/input/plant/greenheart_config_offshore_ca.yaml
```diff
@@ -81,10 +81,10 @@ h2_transport_pipe:
   outlet_pressure: 10 # bar - from example in code from Jamie #TODO check this value
 h2_storage:
   size_capacity_from_demand:
-    flag: False # If True, then storage is sized to provide steady-state storage    
+    flag: True # If True, then storage is sized to provide steady-state storage    
   capacity_from_max_on_turbine_storage: False # if True, then days of storage is ignored and storage capacity is based on how much h2 storage fits on the turbines in the plant using Kottenstete 2003.
-  type: "none" # can be one of ["none", "pipe", "turbine", "pressure_vessel", "salt_cavern", "lined_rock_cavern"]  
-  days: 3 # [days] how many days worth of production we should be able to store (this is ignored if `capacity_from_max_on_turbine_storage` is set to True)
+  type: "lined_rock_cavern" # can be one of ["none", "pipe", "turbine", "pressure_vessel", "salt_cavern", "lined_rock_cavern"]  
+  days: 0 # [days] how many days worth of production we should be able to store (this is ignored if `capacity_from_max_on_turbine_storage` is set to True)
```
jaredthomas68, line 87: Why do some of the yamls have `day: -1` and some have `days: 0`? Do both have the same effect? I think we should be consistent here.

greenheart/tools/eco/hydrogen_mgmt.py
```diff
@@ -355,7 +356,7 @@ def run_h2_storage(
                 area_site,
                 mass_tank_empty_site,
                 _,
-            ) = h2_storage.distributed_storage_vessels(h2_capacity, 1)
+            ) = h2_storage.distributed_storage_vessels(greenheart_config["h2_capacity"], 1)
```
jaredthomas68, line 359: Is there a reason to pull this from the yaml? I think it would make sense to make the capacity choices at the beginning of the function and then use that value.

greenheart/tools/eco/hydrogen_mgmt.py
```diff
@@ -294,6 +298,7 @@ def run_h2_storage(
             hydrogen_storage_capacity_kg, hydrogen_storage_duration_hr, hydrogen_storage_soc = hydrogen_storage_capacity(electrolyzer_physics_results['H2_Results'], greenheart_config['electrolyzer']['rating'], hydrogen_storage_demand)
```
jaredthomas68, line 298: I think it looks like we have two separate sets of logic around sizing, pressurized tower and demand/type. I think we should combine these sets of logic into a single if/else set starting with type for zero storage capacity, then checking demand, from max turbine storage, and finally using hour and max fill rate as default if the others are not specified. Eventually this should have the sizing type as a variable rather than potentially conflicting inputs, but for now we can at least simplify/clarify the logic

greenheart/tools/eco/hydrogen_mgmt.py
jaredthomas68: local variable, though we should save the storage capacity somewhere for output, but not in the config dict.

jaredthomas68: basically leave as it was before and make sure the local variable is getting the right value

greenheart/tools/eco/hydrogen_mgmt.py
```diff
@@ -294,6 +298,7 @@ def run_h2_storage(
...
             greenheart_config["h2_capacity"] = hydrogen_storage_capacity_kg
```
jaredthomas68, line 299: We need to not save results to the config dict because the config dict is persistent across runs that do more than one analysis (optimization, design of experiments, step size studies). Results should only go to results dicts or classes

tests/greenheart/test_greenheart_system.py
```diff
@@ -81,7 +81,7 @@ def test_simulation_wind(subtests):
 
     with subtests.test("lcoh"):
         assert lcoh == approx(
-            7.056407257515502
+            7.248875250895419
```
jaredthomas68, line 84: I'm inclined to use the fixed sizing for these tests and just use the variable sizing in the GreenSteel tests so we avoid changing test values and keep some tests for the fixed storage size approach

tests/greenheart/test_openmdao.py
```diff
@@ -322,15 +322,15 @@ def test_greenheart_component(subtests):
 
     # TODO base this test value on something
     with subtests.test("lcoh"):
-        assert prob["lcoh"][0] == approx(3.04011846, rel=rtol)
+        assert prob["lcoh"][0] == approx(3.1691092704830357, rel=rtol)
```
jaredthomas68, line 325: same comment here. We don't need to worry about variable sizing in the openmdao tests except perhaps in the optimization test to make sure the size is changing as we progress through the optimization.

jaredthomas68 (changes requested):
Looking pretty good. There are some fairly major comments:

1. We need to avoid saving outputs into the config dict or class since the config is persistent across analyses when running an optimization or DOE. 
2. Many test values changing unnecessarily. I think we should just change the GreenSteel tests and the optimization in openmdao tests to use variable sizing and leave the rest as they were.

tests/greenheart/test_greenheart_system.py
kbrunik: These value changes are coming from using `electrolyzer_physics_results['H2_Results']['Time Until Replacement [hrs]']` rather than the `time_between_replacement` for the electrolyzer schedule, not changing hydrogen sizing.

greenheart/tools/eco/hydrogen_mgmt.py
kbrunik: I've changed the logic so that we have a single if/elif/else statement around sizing and then a signle if/elif/else statement around h2 storage technology type. I moved the logic around if on a turbine to inside the specific technology (pressure vessel and turbine) and included a check up front for illegal technology choices based on turbine location selection.

jaredthomas68 (approved):
This looks good to me

greenheart/simulation/greenheart_simulation.py
```diff
@@ -238,38 +239,74 @@ def setup_greenheart_simulation(config: GreenHeartSimulationConfig):
...
+                    "row_spacing": config.greenheart_config["site"]["wind_layout"][
+                        "row_spacing"
+                    ]
+                }
+            )
+            warnings.warn(
+                f"site depth in the orbit_config was {config.orbit_config['plant']['row_spacing']}, but 'row_spacing' in"
+                f"greenheart_config was {config.greenheart_config['site']['wind_layout']['row_spacing']}. The 'row_spacing' value in the orbit_config"
+                "is being overwritten with the value from the greenheart_config",
+                UserWarning,
```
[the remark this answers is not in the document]
bayc: I don't see how the keys changed, just that line breaks were added.

greenheart/tools/eco/electrolysis.py
```diff
@@ -87,6 +88,9 @@ def run_electrolyzer_physics(
         "turndown_ratio": greenheart_config["electrolyzer"]["turndown_ratio"],
     }
 
+    if "time_between_replacement" in greenheart_config['electrolyzer']:
+        warnings.warn("`time_between_replacement` as an input is depricated. It is now calculated internally and is output in electrolyzer_physics_results['H2_Results']['Time Until Replacement [hrs]'].")
```
bayc, line 92: deprecated

tests/greenheart/test_greensteel/input/floris/floris_input_lbw_6MW.yaml
```diff
@@ -0,0 +1,76 @@
+name: Gauss
```
bayc, line 1: This input file is missing the farm section, defining the layout and the turbine type. Is that correct? I don't think FLORIS will run without adding that info.

bayc (changes requested):

tests/greenheart/test_greensteel/input/floris/floris_input_lbw_6MW.yaml
jaredthomas68: GreenHEART overwrites it for now. We would like to have it both ways eventually, but the ORBIT custom layout has not worked for me yet so right now we just use a grid in GreenHEART.

jaredthomas68: So, for GreenHEART we don't have to have that section in the input files.

bayc (approved):

PR #324, opened 2024-06-05
Hydrogen storage sizing

kbrunik:
# steady-state hydrogen storage sizing

Added function to calculate steady-state hydrogen storage capacity based on electrolyzer hydrogen production and hydrogen demand.


jaredthomas68 (approved):
This looks good.

PR #327, opened 2024-06-10
move greensteel tests resource files

jaredthomas68:
# Move GreenSteel test resource files

This PR moves some of the the resource files for testing in the default resource files location internal to hopp.


bayc (approved):
I think these changes are fine. We will want to decided if all testing data should be contained in the examples/other places, or if we want a full set of testing data in the test suite.

kbrunik (approved):

PR #338, opened 2024-08-13
Greensteel-eco-sync

jmartin4nrel:
Adding the ability to set pvwatts losses through the definition of a PVConfig, and thus through the plant input .yaml under "technologies: pv:".

Three types of losses (same text added to the preamble for PVConfig):

dc_ac_ratio: Also known as inverter loading ratio; ratio of max DC output of PV to max AC output of inverter, should be slightly above one (max 1.5) for optimal economics

inv_eff: Inverter efficiency in percent; linear power conversion loss from DC to AC

losses: Any "other" linear power losses in percent, broken down into categories in the GUI version of SAM.


jaredthomas68 (approved):
Looks good to me. May be good to add some tests that make sure everything is being set correctly.

kbrunik (approved):

PR #340, opened 2024-08-22
OM Cost bug

kbrunik:
# Fix bugs in OM modeling

Describe your feature here.


bayc (approved):

greenheart/simulation/greenheart_simulation.py
```diff
@@ -346,7 +346,7 @@ def setup_greenheart_simulation(config: GreenHeartSimulationConfig):
                 )
             ):
                 config.hopp_config["technologies"]["wind"]["fin_model"]["system_costs"][
-                    "om_fixed"
```
bayc, line 349: Should `om_fixed` be used anywhere? Or is it just an incorrect value/not needed for our analyses?

PR #341, opened 2024-08-24
Bug Fix Load Following Dispatch Strategy

kbrunik:
# Bug Fix Load Following Dispatch Strategy

Describe your feature here.


hopp/simulation/technologies/sites/site_info.py
```diff
@@ -66,6 +67,8 @@ class SiteInfo(BaseClass):
     hub_height: hopp_float_type = field(default=97., converter=hopp_float_type)
     capacity_hours: NDArray = field(default=[], converter=converter(bool))
     desired_schedule: NDArrayFloat = field(default=[], converter=converter())
+    curtailment_value_type: str = field(default="desired_schedule", validator=contains(["grid", "desired_schedule"]))
```
genevievestarke, line 70: To be consistent here, we might want to make "grid" the default and have a test to see if a desired schedule is given, otherwise if it's set to desired_schedule and site.py doesn't have that variable, you'll get an error

genevievestarke (changes requested):
Looks good, just might want to touch base on what we want the default of this parameter to be

hopp/simulation/technologies/sites/site_info.py
genevievestarke: Never mind, I just realized it's in the desired_schedule loop so you won't get an error. I still think we should put grid as the default though since it seems to be what people prefer

kbrunik: Okay I changed it to default to "grid"

bayc (approved):

hopp/simulation/technologies/grid.py
```diff
@@ -118,29 +118,46 @@ def simulate_grid_connection(
...
-                x * 1e3 for x in self.site.desired_schedule],
-                int(project_life / (len(self.site.desired_schedule) // self.site.n_timesteps))
-            )
-            self.generation_profile = list(np.minimum(total_gen, lifetime_schedule)) # TODO: remove list() cast once parent class uses numpy 
-
-            self.missed_load = np.array([schedule - gen if gen > 0 else schedule for (schedule, gen) in
-                                     zip(lifetime_schedule, self.generation_profile)])
-            self.missed_load_percentage = sum(self.missed_load)/sum(lifetime_schedule)
-
+            if self.site.curtailment_value_type == "grid":
```
jaredthomas68, line 121: why is this type called "grid"?

hopp/simulation/technologies/grid.py
```diff
@@ -118,29 +118,46 @@ def simulate_grid_connection(
...
+                lifetime_schedule: NDArrayFloat = np.tile([self.interconnect_kw],
```
jaredthomas68, line 122: Why call this "lifetime_schedule"? Isn't this the maximum allowed generation?

jaredthomas68: I guess it is legacy

hopp/simulation/technologies/grid.py
jaredthomas68: Let's change this to `interconnect_kw`

jaredthomas68 (changes requested):

genevievestarke (approved):

hopp/simulation/technologies/grid.py
jaredthomas68: Thanks for the correction. I think this is much more clear.

jaredthomas68 (approved):

PR #345, opened 2024-09-03
Bugfix: Remove unused property that could cause recursion loop

bayc:
# Remove cost property that can cause recursion loop when accessed

This PR removes the `om_total_expense` property from the custom financial model. This bug was identified and fixed by @elenya-grant when using the python debugger.

## Related issue
None.

`simulation/technologies/financial/custom_financial_model.py`

None.

Tests are passing.

### DIFF
--- a/hopp/simulation/technologies/financial/custom_financial_model.py
+++ b/hopp/simulation/technologies/financial/custom_financial_model.py
@@ -393,9 +393,10 @@ class CustomFinancialModel():
     def annual_energy(self) -> float:
         return self.value('annual_energy_pre_curtailment_ac')
     
-    @property
-    def om_total_expense(self) -> float:
-        return self.value('om_total_expense')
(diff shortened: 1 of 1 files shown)


hopp/simulation/technologies/financial/custom_financial_model.py
```diff
@@ -393,9 +393,10 @@ def export_battery_values(self):
     def annual_energy(self) -> float:
         return self.value('annual_energy_pre_curtailment_ac')
     
-    @property
-    def om_total_expense(self) -> float:
-        return self.value('om_total_expense')
+    #below throws a recursion error in debug mode
+    # @property
+    # def om_total_expense(self) -> float:
+    #     return self.value('om_total_expense')
```
dguittet, line 399: If you replaced this with something like

`return self.Outputs. om_total_expense`

would that work?

PR #346, opened 2024-09-03
Bugfix: correctly set battery O&M costs

bayc:
# Update process of setting battery O&M costs

This PR removes the optional cases for setting the battery O&M costs as they were not being updated properly when using the custom financial model. This bug was identified and corrected by @elenya-grant.

## Related issue
None.

`hopp/simulation/technologies/power_source.py`

None.

Tests are passing locally.

### DIFF
--- a/hopp/simulation/technologies/power_source.py
+++ b/hopp/simulation/technologies/power_source.py
@@ -449,18 +449,18 @@ class PowerSource(BaseClass):
     @property
     def om_capacity(self):
         """Capacity-based O&M amount [$/kWcap]"""
-        if self.name != "Battery":
-            return self._financial_model.value("om_capacity")
-        return self._financial_model.value("om_batt_capacity_cost")
+        # if self.name != "Battery":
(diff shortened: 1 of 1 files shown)


bayc:
> I think we need to make updates to the `CustomFinancialModel` to make sure it's updating properly rather than deleting these if statements. `test_hybrid.py::test_hybrid_om_costs` is failing and that means that if we pull in this change that you won't be able to use `SingleOwner`.

@kbrunik Was this addressed by #361?

kbrunik:
> > I think we need to make updates to the `CustomFinancialModel` to make sure it's updating properly rather than deleting these if statements. `test_hybrid.py::test_hybrid_om_costs` is failing and that means that if we pull in this change that you won't be able to use `SingleOwner`.
> 
> @kbrunik Was this addressed by #361?

No, this could be converted to an issue and this PR closed because I think it plays into a larger question of how to support the custom financial model.

kbrunik:
Created issue since this is likely a larger discussion.

PR #350, opened 2024-09-26
Bug_Fix: Property Recursion

kbrunik:
# Update cost property that was causing a recursion loop when accessed

This PR is an update to #345 based on reviewer feedback. 

- Renamed property `om_total_expense` to `om_total_expenses` following naming convention in `HybridSimulation`. 
- Updated type `om_total_expense` in the `Outputs` class and the `om_total_expenses` property to a Sequence.
- Changed `cf_operating_expenses` from a Numpy Array type to a Tuple to avoid additional errors when accessing the property.
- No recursion loop occurred when accessing the property after these updates.

### DIFF
--- a/hopp/simulation/technologies/financial/custom_financial_model.py
+++ b/hopp/simulation/technologies/financial/custom_financial_model.py
@@ -393,9 +393,10 @@ class CustomFinancialModel():
     def annual_energy(self) -> float:
         return self.value('annual_energy_pre_curtailment_ac')
     
-    @property
-    def om_total_expense(self) -> float:
-        return self.value('om_total_expense')
+    #below throws a recursion error in debug mode
+    # @property
+    # def om_total_expense(self) -> float:
+    #     return self.value('om_total_expense')
     
     # for compatibility with calls to SingleOwner
     @property
--- a/hopp/simulation/technologies/financial/custom_financial_model.py
+++ b/hopp/simulation/technologies/financial/custom_financial_model.py
@@ -393,10 +393,9 @@ class CustomFinancialModel():
     def annual_energy(self) -> float:
         return self.value('annual_energy_pre_curtailment_ac')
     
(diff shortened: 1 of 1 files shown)


hopp/simulation/technologies/financial/custom_financial_model.py
```diff
@@ -395,7 +395,7 @@ def annual_energy(self) -> float:
     
     @property
     def om_total_expense(self) -> float:
-        return self.value('om_total_expense')
+        return self.Outputs.om_total_expense
```
jaredthomas68, line 398: Where are values being assigned to the Outputs class?

jaredthomas68 (commented):
I think I need to understand how this works better. I don't see any way that the items in Outputs are being assigned.

hopp/simulation/technologies/financial/custom_financial_model.py
kbrunik: That's a good question. Upon further investigation it doesn't get assigned. I mirrored the structure in the `lcoe_real` property since the naming convention switches between `HybridSimulation` and `PowerSource` and `CustomFinancialModel` so that it returns `om_total_expense` calculated in `PowerSource`

dguittet: It's true that there's a lot of functionality that hasn't been implemented in this class yet.

dguittet (approved):

hopp/simulation/technologies/financial/custom_financial_model.py
kbrunik: So for the Outputs class, not all of the attributes in that class are being assigned. Think of it as the framework (based on PySAM nomenclature) that can be filled in as we need.

dguittet: Exactly!

PR #352, opened 2024-10-01
Dev/refactor

jmartin4nrel:
# Add meaningful title here

Describe your feature here.


jmartin4nrel:
Test appears to be failing due to a deprecated numpy call in ORBIT code, idk how to handle that...

RHammond2:
@jmartin4nrel the easiest way to resolve this is to set the "numpy<2" in the `requirements.txt`, which works on my end for both your updates and the base branch. That said this should be resolved with ORBIT v1.1, but that's not relevant to this PR and should be addressed separately. I believe any existing environments would have to be updated with `pip install --upgrade numpy<2` in order to work.

jaredthomas68 (approved):
Thanks for the fix. Looks good to me.

PR #359, opened 2024-10-14
Enhancement: Upgrade Python Infrastrucutre and Compatibility

RHammond2:
# Convert to Modern Python Management

This PR converts the multi-file (`setup.py`, `requirements.txt`, and etc.) to a `pyproject.toml` workflow and updates the allowable Python versions to 3.9, 3.10, and 3.11 (max allowed) in support of the HOPP v3 release.

## Related issue

Closes #354 

## Impacted areas of the software

- All installation files (`setup.py`, `requirements.txt`, `requirements-dev.txt`, `hopp/version.py`, and `hopp/__init__.py` have been removed or updated to add in the singular `pyproject.toml`
- The GitHub Actions have all been updated to their modern workflow versions and to rely on Python 3.9
- `README.md` has been updated to highlight the new installation usage

## Additional supporting information

N/A

## Test results, if applicable

N/A, no changed model logic.

PR #362, opened 2024-10-15
Enhancement: Convert to PyPI Trusted Publshing

RHammond2:
# Use PyPI Trusted Publishing and Test-PyPI Runner

This PR adds updates the PyPI workflow to use the trusted publisher workflow recommended by PyPI. In addition, this adds in a Test PyPI runner that allows for a separate tag creation & push step to check that the package builds prior to creating a release.

## Related issue

N/A

- `publish_to_pypi.yml`: build steps are modified to use the new workflow
- `publish_to_test_pypi.yml`: added 
- `README.md`: adds PyPI badges to highlight latest version and supported Python
- `pyproject.toml`: updates keywords to be matching the project

The reason behind the switch is to modernize the publishing workflow to maximize package security and streamline the workflows.

N/A

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -7,21 +7,26 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    if: github.event.action == 'published'
+    environment:
+      name: PyPI Publishing
+      url: https://pypi.org/p/HOPP
+    permissions:
+      id-token: write
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
         python-version: 3.9
-    - name: Install dependencies
+    - name: Build package
       run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine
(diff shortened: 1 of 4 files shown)


johnjasa:
@RHammond2, do you know why the RTD build would fail? I see it's not just for this PR but also for others (e.g. #356). Is it caused by the bump from 3.8 to 3.9?

RHammond2:
That should've been handled in #359, so I'm not sure what happened to break compatibility because it worked prior to merging updates in from develop, though nothing I reconfigured previously has been changed. I'll try rerunning it and see what happens because it just fails to checkout at the latest commit.

johnjasa:
> @johnjasa, that should've been handled in #359, so I'm not sure what happened to break compatibility because it worked prior to merging updates in from develop, though nothing I reconfigured previously has been changed. I'll try rerunning it and see what happens because it just fails to checkout at the latest commit.

Gotcha, thanks! I'll acknowledge that I merged develop into this PR due to upstream changes -- maybe that affected the RTD build?

johnjasa (approved):

PR #363, opened 2024-10-16
Test floris

jaredthomas68:
# Add tests for FLORIS

This PR adds tests for the FLORIS capabilities in HOPP. The tests do not have meaningful values other than what I got when running the code locally. This PR also updates the hybrid tests for wind only with a new tolerance because the existing tolerance was so large as to be meaningless. 

## Related issue

#357 

`tests/hopp/test_wind.py`
`tests/hopp/test_hybrid.py`

This PR is a critical step towards updating the FLORIS version used by HOPP.

All tests are passing locally.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
     "Pyomo>=6.1.2",
     "diskcache",
     "fastkml",
-    "floris",
+    "floris<=3.6",
     "future",
     "global_land_mask",
     "humpday",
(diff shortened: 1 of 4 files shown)


johnjasa (approved):

johnjasa (changes requested):
Looks like we might be missing a floris config file that you might have locally as part of this commit; could you please update this PR? https://github.com/NREL/HOPP/actions/runs/11357651664/job/31590877596?pr=363#step:6:230

johnjasa (approved):

PR #377, opened 2024-10-22
make battery heuristic general so other generation tech besides wind and pv can use the heuristic

jaredthomas68:
…and pv can be used

# Generalize heuristic load following to any gen tech

This PR allows any technology to be dispatched with the load following heuristic, not just wind and pv.

`hopp/simulation/technologies/dispatch/hybrid_dispatch_builder_solver.py`

This came up while working on the profast financial model and needs to get merged in prior to the financial model work

Passing


genevievestarke (approved):
Looks good to merge to me!

hopp/simulation/technologies/dispatch/hybrid_dispatch_builder_solver.py
```diff
@@ -674,12 +674,12 @@ def simulate_with_dispatch(
 
     def battery_heuristic(self):
         tot_gen = [0.0] * self.options.n_look_ahead_periods
```
RHammond2, line 676: Could this be simplified to the following?

```python
tot_gen = np.zeros(self.options.n_look_ahead_periods)
for power_source in self.power_sources.keys():
    if "battery" in power_source or "grid" in power_source:
        continue
    tot_gen += self.power_sources[power_source].dispatch.available_generation
```

jaredthomas68: Yes and done

tests/hopp/test_dispatch.py
```diff
@@ -1023,4 +1044,52 @@ def test_hybrid_dispatch_baseload_heuristic_and_analysis(site):
...
+def test_dispatch_load_following_heuristic_with_wave(site, subtests):
+    dispatch_options = {'battery_dispatch': 'load_following_heuristic', 'grid_charging': False}
+    wave_battery = {key: technologies[key] for key in ['wave', 'battery', 'grid']}
+
+    for tech in wave_battery.keys():
+        wave_battery[tech]["fin_model"] = default_fin_config
+
+    wave_resource_file = ROOT_DIR / "simulation" / "resource_files" / "wave" / "Wave_resource_timeseries.csv"
+
+    desired_schedule=desired_schedule = 8760*[20]
```
RHammond2, line 1058: Looks like there is an extra `desired_schedule=` here.

hopp/simulation/technologies/dispatch/hybrid_dispatch_builder_solver.py
RHammond2: I think it needs to get pushed, but I'll trust it's on its way.

tests/hopp/test_dispatch.py
jaredthomas68: Thanks, fixed

hopp/simulation/technologies/dispatch/hybrid_dispatch_builder_solver.py
jaredthomas68: Good catch. I'm not sure lumpy will work without having knock-on effects, but I'll try it

PR #383, opened 2024-10-23
Remove unused dependencies

### LINKED ISSUE #303: Consider using just the yaml package for handling !include funtionality
# Use yaml package for !include functionality
Currently, we depend on `pyyaml-include` package for enabling `!include` functionality in our input files. This feature would remove that dependency and adopt the same method that is used in the FLORIS repository for supporting the `!include` functionality.

## Proposed solution
We would implement the same strategy as used in FLORIS, shown here: https://github.com/NREL/floris/blob/main/floris/utilities.py#L340

## Alternatives considered
Keep it the way it is, but currently running into testing issues after `pyyaml-include` released a new major version.

bayc:
# Remove unused dependencies

This PR just removes currently unused dependencies. These were used for code in the `alt_dev` directory, developed by @qualand, who has approved removal of them for now and asked that we consider integrating them into future development. Also removed `pyyaml-include` as we use the `load_yaml` utility function now that uses the standard `yaml` package.

## Related issue

Closes #303 

`pyproject.toml`

None.

Tests were passing on my fork. Looking into the failing tests now. Seems readthedocs is failing because of missing packages for the `alt_dev` directory. Confirming with @dguittet whether that directory is still needed or not.


pyproject.toml
```diff
@@ -15,29 +15,23 @@ dependencies = [
     "NREL-PySAM==4.2.0",
     "Pillow",
     "Pyomo>=6.1.2",
-    "diskcache",
```
RHammond2, line 18: The docs build is failing because `alt_dev` requires this package. Was this just an unused import, or is it actually used in the model?

RHammond2 (approved):
This looks good to me, though I'm a bit unfamiliar with FLORIS 4.0, so I'm trusting the implementation is corrected now.

pyproject.toml
bayc: Yea, just noticed that as well. It appears to be used in `alt_dev`, which I imagine several of the other packages were also used in. I've reached out to @dguittet to see if this directory should be kept around or not. Looks like the last development was 2 years ago (we moved some resource files 5 months ago, but that was done to the whole repo).

RHammond2:
> This looks good to me, though I'm a bit unfamiliar with FLORIS 4.0, so I'm trusting the implementation is corrected now.

No idea how I commented on the wrong PR, please ignore this comment.

johnjasa (approved):

pyproject.toml
bayc: Confirmed with @qualand that the directory could be removed for the next release.

PR #391, opened 2024-10-25
Update HOPP Examples

kbrunik:
# Updating HOPP examples to HOPP V3.0

- Remove PEM electrolyzer example.
- Update FLORIS example to be compatible with FLORIS V4.
- Changed battery size to be a 20MW battery with 4 hr duration (instead of the reverse).

## Related issue
Made an issue #393 because there were Warnings when running `06-wave-wind.ipynb`.


johnjasa (approved):
Just pushed some small changes directly to the branch, this looks great! Thank you for updating these.

PR #405, opened 2024-11-13
Feature: PV Tilt Angle

kbrunik:
# Feature: PV Tilt Angle

#### Feature added allows the user to set the PV panel tilt angle in the HOPP config YAML.
- Panel tilt angle can be specified by either setting a string or a float in the config YAML. If not set the `panel_tilt_angle` property defaults to using a latitude function to calculate the optimal tilt angle for the panel based on the site location latitude.
- If a float is in the config the `__attrs__post_init__` in `PVPlant` will ensure the value is between 0 and 90 degrees (exclusive) and set the `panel_tilt_angle` property to the float value provided.

- If a str is in the config the  `__attrs__post_init__` in `PVPlant` will:
1) if the str is"lat", set the `panel_tilt angle` property based on the latitude in the site definition.
2) if the str is "lat-func", set the `panel_tilt angle` property based on calculating the optimal tilt angle based on information from this [website](https://www.solarpaneltilt.com/).

`PVPlant.py`

Updated tests that include the PVPlant, they are failing because the default is to use the "lat-func" to calculate the panel tilt angle whereas previously the `panel_tilt_angle` was not being set and defaulted to 0. 

Added tests to `test_pv_source.py` to test all options to set `panel_tilt_angle`, the TypeErrors, and Exceptions.


hopp/simulation/technologies/pv/pv_plant.py
```diff
@@ -220,6 +250,18 @@ def footprint_area(self):
         area = num_modules * module_attribs['area']
         return  area
 
+    @property
+    def land_coverage_area(self):
```
elenya-grant, line 254: We talked about minor changes to the ``land_coverage_area`` property. I think it's fine to leave as-is and we can revisit it later OR to remove this property for now. What we discussed were
1) naming of ``land_coverage_area`` vs ``footprint_area``
2) I think the land-coverage area calc needs to include ground-coverage ratio (``self._system_model.SystemDesign.gcr``). 

These changes are not required for approval.

elenya-grant (approved):
looks good to me!

hopp/simulation/technologies/pv/pv_plant.py
kbrunik: How would you update the calculation to include the ground-coverage ratio?

hopp/simulation/technologies/pv/pv_plant.py
```diff
@@ -73,6 +78,30 @@ class PVPlant(PowerSource):
...
+                    tilt = self.site.lat*0.87
+                elif self.site.lat>25 and self.site.lat<=50:
+                    tilt = (self.site.lat*0.76) + 3.1
+                else:
+                    tilt = self.site.lat
+            else:
+                raise Exception("The str can be set to 'lat' or 'lat-func'.")
+                
+        elif isinstance(self.config.panel_tilt_angle,float):
+            if self.config.panel_tilt_angle > 0 and self.config.panel_tilt_angle<90:
```
jaredthomas68, line 96: Is there a reason to exclude 0 and 90?

hopp/simulation/technologies/pv/pv_plant.py
```diff
@@ -220,6 +254,18 @@ def footprint_area(self):
         area = num_modules * module_attribs['area']
         return  area
 
+    @property
+    def land_coverage_area(self):
+        """Estimate Total Module Land Cover Area [m^2]"""
+        import numpy as np
```
jaredthomas68, line 260: I think we should move the import to the top of the file

hopp/simulation/technologies/pv/pv_plant.py
```diff
@@ -220,6 +254,18 @@ def footprint_area(self):
...
+        module_attribs = get_module_attribs(self._system_model)
+        tilt_angle_max_deg = self._system_model.value("tilt") + (self._system_model.value("rotlim")/2)
+        # tilt_angle_min_deg = self._system_model.value("tilt") - (self._system_model.value("rotlim")/2)
+        max_land_cover_module_m2 = module_attribs['width']*(module_attribs['length']*np.cos(np.deg2rad(tilt_angle_max_deg)))
+        num_modules = self.system_capacity_kw / module_attribs['P_mp_ref']
+        land_coverage_area = num_modules * max_land_cover_module_m2 #m^2
+        return land_coverage_area
```
jaredthomas68, line 267: I'm inclined to save this property for a separate PR.

tests/hopp/test_hybrid.py
```diff
@@ -1392,7 +1438,7 @@ def reinstate_orig_values():
     )
 
     capcred = hybrid_plant.capacity_credit_percent
-    assert capcred["pv"][0] == approx(8.03, rel=0.05)
+    assert capcred["pv"][0] == approx(6.85, rel=0.05)
```
jaredthomas68, line 1441: I don't understand why this would decrease, but perhaps I do not understand what it is.

jaredthomas68 (changes requested):
Looking pretty good. A few comments. Thanks for putting this together

hopp/simulation/technologies/pv/pv_plant.py
kbrunik: Same, I've removed the commit that added the property.

hopp/simulation/technologies/pv/pv_plant.py
kbrunik: Not it doesn't appear that PySAM is exclusive so I've updated it to include 0 and 90.

hopp/simulation/technologies/pv/pv_plant.py
kbrunik: removed property

tests/hopp/test_hybrid.py
kbrunik: This is where the capacity credit percent is calculated, it looks to me that it should increase given increased AEP, so it does seem strange. https://github.com/NREL/HOPP/blob/53490cb54abfe024e6338310a88374707d54ed20/hopp/simulation/technologies/power_source.py#L210

jaredthomas68: Thanks for that. I wonder if the problem is in changes made to the `interconnect_kw` for dispatching. I have not investigated, just throwing that out for now.

johnjasa (approved):
Thanks for this PR and for adding the permutations of tests! Great addition.

jaredthomas68 (approved):
We need to follow up with #406 still, but let's go ahead and pull this in.

PR #380, opened 2024-10-23
Cambium API and GREET Excel Integration

dakotaramos:
Cambium and GREET integration
Added:
cambium_data.py: to manage the Cambium API call and parsing of data
greet_data.py: to handle manipulation and extraction of data from GREET excel docs
resource_files/cambium/: directory that stores cambium resource files, files currently in this folder are defaults for testing
resource_files/greet/: directory that stores 2023 versions of GREET, a preprocessed yaml file with relevant data points to be used in GreenHEART LCA analysis, and associated README.md
tests/hopp/test_cambium_resource.py: pytests for cambium integration
tests/hopp/test_greet_resource.py: pytests for greet integration

Edited:
tests/hopp/utils.py: created default file paths for cambium and greet testing
hopp/simulation/technologies/resource/__init__.py: made cambium and greet data classes importable
pyproject.toml: added "pytest-dependency" under develop dependencies, used in cambium pytests

## Related issue
https://github.com/NREL/HOPP/issues/375

## Impacted areas of the software
pyproject.toml, pytest suite, and resources

## Additional supporting information
This work is being completed for the LCA integration, the data pulling and parsing will live in HOPP, the LCA calculations with that data will live in GreenHEART

NOTE: 

- Additional variables can possibly be added to the Cambium API call to pull curtailment, end use costs, busbar costs, clean energy fractions, distribution losses, transmissions totals, and various load values if needed. 

- Spoke with Cambium API folks (James Morris). 1) they do not have any plans in FY25 to add a bulk requests functionality to the API but will let us know when it becomes available. 2) Communicated possible increased API requests from this integration so they can monitor system health / computational resources needed.

- Many more data points can be extracted from GREET excel, but requires time and effort to interpret GREET calculations and determine proper values.

## Test results, if applicable
tests/hopp/test_cambium_resource.py ..... 
tests/hopp/test_greet_resource.py ...

PR #408, opened 2024-12-04
GREET Data Hotfix

dakotaramos:
Hotfix to correct values and units for iron ore mining and pelletizing emission intensities based on feedback from ANL.

## Related issue
https://github.com/NREL/HOPP/pull/380

resources

This work is being completed for the LCA integration, the data pulling and parsing will live in HOPP, the LCA calculations with that data will live in GreenHEART

tests/hopp/test_greet_resource.py ...

### DIFF
--- a/hopp/simulation/resource_files/greet/2023/greet_2023_processed.yaml
+++ b/hopp/simulation/resource_files/greet/2023/greet_2023_processed.yaml
@@ -1,5 +1,7 @@
-DRI_iron_ore_mining_EI: 0.22901198723003552
-DRI_iron_ore_pelletizing_EI: 0.23105675045934587
+DRI_iron_ore_mining_EI_per_MT_ore: 125.60298098051585
+DRI_iron_ore_mining_EI_per_MT_steel: 229.01198723003554
+DRI_iron_ore_pelletizing_EI_per_MT_ore: 126.7244434860691
+DRI_iron_ore_pelletizing_EI_per_MT_steel: 231.0567504593459
 NG_combust_EI: 56.531234265084805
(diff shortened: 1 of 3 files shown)


johnjasa:
Thanks for this, @dakotaramos! The changes you've made look good.
The failing CI appears unrelated to your changes. @RHammond2, could you please take a look and see if a package update or similar has caused these tests to fail?

RHammond2:
> Thanks for this, @dakotaramos! The changes you've made look good. The failing CI appears unrelated to your changes. @RHammond2, could you please take a look and see if a package update or similar has caused these tests to fail?

The issue is stemming from the `KML` object having `features` as a callable, so if we change [these lines](https://github.com/dakotaramos/HOPP/blob/8a3b94a430c91def08c7b75d676b07bf373a9cb4/hopp/simulation/technologies/sites/site_info.py#L253-L254) to use `k.features` instead of `k.features()`, then we should be ok in theory.

However, something I noticed is that we'll also be attempting to index an empty list in the tests, so this will raise a new error. Doubly however, this is call was not previously an issue, so I checked out the `fastkml` changelog, and they had a v1.0 release in mid-November.

tl;dr: Pin fastkml to be < 1, and open an issue to migrate HOPP and GreenHEART to be compatible with fastkml v1.0.

RHammond2 (approved):
Aside from the `fastkml` issue in my comment this all looks good to me.

dakotaramos:
@RHammond2 @johnjasa just pushed the update to the fastkml version in pyproject.toml

johnjasa (approved):

Issue #428, opened 2025-02-10
Task: Distributed Wind Examples


elenya-grant:
# Add distributed wind examples

Requires the following features/bug-fixes:

- [x] add option to specify different locations and/or resource years for wind and solar resource
- [/] simplify defining site boundaries in SiteInfo.py
- [/] add ability to adjust wind air density based on site elevation
- [ ] add simple-grid turbine layout option
- [ ] integrate turbine layout with floris.py
- [ ] add auto-generation of wind farm layout for floris.py
- [ ] integrate turbine-models library
- [ ] (optional) update floris.py for simulating multiple turbine types
- [ ] make examples highlighting these developments
- [ ] add tests for floris usage

## Proposed solution
all of these capabilities exist in https://github.com/elenya-grant/HOPP/tree/v3/turb_tools

## Additional context
Need to have merged in by end of Feb 2025 to show example at a conference.

PR #435, opened 2025-02-20
Feature add: integration with turbine-models library for wind simulations

elenya-grant:
# Turbine-models library integration!

Integrate the ability to simulate wind turbines available in the [turbine-models package](https://github.com/NREL/turbine-models/tree/master) with both PySAM and FLORIS wind simulation options!

## PR Checklist

- [x] Documentation
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
  - [-] Examples have been updated

## Related issues

Issue #428 

### Modified workflow files and functions
- `hopp/simulation/technologies/wind/wind_plant.py`
    - `WindConfig`:
        - added attributes: `turbine_name` and `turbine_management`
    - `WindPlant`:
        - `__attrs_post_init__()`: updated to call `initialize_pysam_wind_turbine()` if using pysam
        - `initialize_pysam_wind_turbine()`: new function to handle pysam-specific attribute setting, set turbine model from turbine-models library, and redownload resource data if discrepancy in turbine-height and wind resource data heights.
        - `num_turbines` setter: modified so if using "custom" layout, it doesnt throw user-warning upon initialization (which would happen previously when called from `__attrs_post_init__() method`)
- `hopp/simulation/technologies/wind/floris.py` class `Floris`
    - `initialize_from_floris()`: modified to call `initialize_wind_turbine()` and to re-download resource data if discrepancy in turbine hub-height and wind resource data heights.
    - `initialize_wind_turbine()`: new function to set turbine model in `floris_config` upon Floris initialization
    - `update_wind_turbine()`: new function to update turbine model after Floris has been initialized.
- `hopp/tools/resource/wind_tools.py`
    - `parse_resource_data`: modified to not average resource data for multiple heights if hub-height is equal to lower or upper resource height.

### New files and functions
- `hopp/tools/design/wind/`: new files and functions
    - `floris_helper_tools.py`
        - `check_output_formatting()`: recursive method to convert types in a dictionary to human readable format if its exported to a yaml file
        - `write_floris_layout_to_file()`: function to write a wind farm layout to FLORIS-compatible yaml file format.
        - `write_turbine_to_floris_file()`: function to write turbine model to FLORIS-compatible yaml file format.
        - `check_floris_library_for_turbine()`: returns true if turbine_name is in the FLORIS internal turbine library, otherwise returns false
        - `load_turbine_from_floris_library()`: loads a turbine from the FLORIS internal turbine library.
        - `check_libraries_for_turbine_name_floris()`: checks FLORIS internal turbine library and turbine-models library for turbine name, returns a FLORIS-compatible turbine model dictionary if its a valid turbine name, returns a string if the turbine is not found that can be used in a UserWarning message.
    - `power_curve_tools.py`: 
        - `plot_power_curve()`: plot Cp and Ct curve across wind speeds (created by @genevievestarke)
        - `pad_power_curve()`: if a power-curve is only defined from cut-in to cut-out wind speeds, this pads the power-curve from 0 m/s to 50 m/s wind speed so no errors are thrown in wind simulations.
        - `calculate_cp_from_power()`: Calculate power coefficient curve (Cp) from power curve.
        - `calculate_power_from_cp()`: Calculate power curve from power coefficient curve (Cp).
        - `estimate_thrust_coefficient()`: Calculate thrust coefficient curve (Ct) from power coefficient curve (Cp). (created by @genevievestarke)
    - `turbine_library_interface_tools.py`
        - `extract_power_curve()`: gets Power, Cp, and Ct curve for turbine and formats for simulations.
        - `check_hub_height()`: if turbine from turbine-models library has multiple hub-height options, checks the other places a user may have specified a specific hub-height or sets to median hub-height.
        - `get_pysam_turbine_specs()`: Load turbine data from turbine-models library in PySAM-compatible format.
        - `get_floris_turbine_specs()`: Load turbine data from turbine-models library in FLORIS-compatible format.
    - `turbine_library_tools.py` integrated functions:
        - `check_turbine_name()`: returns a valid turbine name in turbine-library that is best-match to user-provided turbine name.
        - `check_turbine_library_for_turbine()`: returns True if turbine is found in turbine library, returns false otherwise

### New tests

All new tests are in `tests/hopp/`

- `test_wind_design_tools.py`: unit tests for floris integration tools
    - `test_floris_library_tools_for_valid_floris_turbine()`: unit test for `check_floris_library_for_turbine` and `load_turbine_from_floris_library` in `floris_hepler_tools` with a valid floris-library turbine name
    - `test_floris_library_tools_for_invalid_floris_turbine()`: unit test for `check_floris_library_for_turbine` and `load_turbine_from_floris_library` in `floris_hepler_tools` with an invalid floris-library turbine name
    - `test_floris_turbine_loader_valid_floris_turbine()`: unit test for `check_libraries_for_turbine_name_floris()` in `floris_hepler_tools`
    - `test_floris_turbine_loader_multi_hub_height_turbine()`: functions as a unit test for `check_hub_height()` in `turbine_library_interface_tools.py`
    - `test_floris_turbine_loader_single_hub_height_turbine()`: functions as a unit test for `check_hub_height()` in `turbine_library_interface_tools.py`
- `test_turbine_models_interface.py`
    - `test_turbine_library_tools_for_valid_turbine_name()`: unit test for `check_turbine_name` and `check_turbine_library_for_turbine` in `turbine_library_tools.py`
    - `test_turbine_library_tools_for_invalid_turbine_name()`: unit test for `check_turbine_name` and `check_turbine_library_for_turbine` in `turbine_library_tools.py`
    - `test_floris_nrel_5mw()`: tests WindPlant integration for runnning floris with a turbine from the FLORIS internal turbine library
    - `test_floris_nrel_5mw_hopp()`: tests HOPP integration for runnning floris with a turbine from the FLORIS internal turbine library
    - `test_floris_NREL_5MW_RWT_corrected_hopp()`: tests HOPP integration for runnning floris with a turbine from the turbine-models library.
    - `test_pysam_NREL_5MW_RWT_corrected_hopp()`: tests HOPP integration for runnning pysam with a turbine from the turbine-models library. 
        
### Other files modified:
- `pyproject.toml`: modified to include turbine models library in dependencies
- `RELEASE.md`: added in feature info
- `README.md`: changed language to include "distributed"


hopp/simulation/technologies/layout/wind_layout.py
```diff
@@ -435,7 +415,6 @@ def reset_grid(self, n_turbines):
     def reset_basic_grid(self,n_turbines):
         """Create a most-square `basicgrid` layout for specified number of turbines.
         requires parameters are `WindBasicGridParameters`.
-
```
RHammond2, line 438: Can you add this blank line back?

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -191,17 +199,68 @@ def __attrs_post_init__(self):
...
+            else:
+                turbine_name = self.config.turbine_name
+            turbine_dict = get_pysam_turbine_specs(turbine_name,self)
+            self._system_model.Turbine.assign(turbine_dict)
+            self.rotor_diameter = turbine_dict["wind_turbine_rotor_diameter"]
+            self.turb_rating = max(turbine_dict["wind_turbine_powercurve_powerout"])
+
+            if self.config.rotor_diameter is not None:
+                if self.config.rotor_diameter != self._system_model.Turbine.wind_turbine_rotor_diameter:
+                    raise UserWarning(f"input rotor diameter ({self.config.rotor_diameter}) does not match rotor diameter for turbine ({self._system_model.Turbine.wind_turbine_rotor_diameter})")
```
RHammond2, line 238: Can you split this across multiple lines? I'm not aiming for a specific line length but once it gets past well beyond 122 I think it's safe to break up as a good practice. There are a few more here that would be nice as well. I recommend the below form.

```python
msg = (
    "message"
    " continued"
)
raise UserWarning(msg)
```

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -191,17 +199,68 @@ def __attrs_post_init__(self):
...
+        
+        if self.config.hub_height is not None:
+            if self.config.hub_height != self._system_model.Turbine.wind_turbine_hub_ht:
+                raise UserWarning(f"input hub-height ({self.config.hub_height}) does not match hub-height from for turbine ({self._system_model.Turbine.wind_turbine_hub_ht})")
+        
+        if self._system_model.Turbine.wind_turbine_hub_ht != self.site.wind_resource.hub_height_meters:
+            if self._system_model.Turbine.wind_turbine_hub_ht >= min(self.site.wind_resource.data["heights"]) and self._system_model.Turbine.wind_turbine_hub_ht<=max(self.site.wind_resource.data["heights"]):
```
RHammond2, line 245: Same here, but giving each check its own line becomes preferred so it reads in a self-documenting style.

```python
valid_min_height = ...  # left-side check
valid_max_height = ...  # right-side check
if valid_min_height and valid_max_height:
    ....  # rest of code
```
```

File: hopp/simulation/technologies/wind/floris.py [hunk 4]
```diff
@@ -101,61 +101,69 @@ def __attrs_post_init__(self):
...
-                f"num_turbines in WindConfig ({self.config.num_turbines}) does not equal "
-                f"number of turbines in floris config layout ({self.nTurbs})"
-            )
+       
         return floris_config
-
+    
     def value(self, name: str, set_value=None):
-        """Set or retrieve attribute of `hopp.simulation.technologies.wind.floris.Floris`.
+        """Set or retrieve atrribute of `hopp.simulation.technologies.wind.floris.Floris`.
```
RHammond2, line 166: "attribute"

hopp/simulation/technologies/wind/floris.py
```diff
@@ -168,15 +176,35 @@ def value(self, name: str, set_value=None):
         else:
             return self.__getattribute__(name)
 
-    def set_floris_value(self, name, value):
+    def set_floris_value(self,name,value):
```
RHammond2, line 179: Can you add a space after each comma, and in the next few methods?

hopp/simulation/technologies/wind/floris.py
```diff
@@ -168,15 +176,35 @@ def value(self, name: str, set_value=None):
...
+        """
         if value is not None:
             self.fi.set(**{name:value})
+    
+    def set_floris_param(self,param,value):
+        """Set parameter of FlorisModel object using the `set_param` function.
 
-    def set_floris_param(self, param, value):
+        Args:
+            param (str): name of parameter to update.
```
RHammond2, line 193: This should align with the FLORIS `set_param`, which takes a list of strings, so it's not just a single name, unless it's a length-1 list.

hopp/simulation/technologies/wind/floris.py
```diff
@@ -168,15 +176,35 @@ def value(self, name: str, set_value=None):
...
+        """
         if value is not None:
-            self.fi.set_param(param, value)
+            self.fi.set_param(param,value)
+    
+    def get_floris_param(self,param):
+        """Get parameter of FlorisModel object using the `get_param` function.
+
+        Args:
+            param (str): name of parameter in FlorisModel object.
```
RHammond2, line 203: Same comment here on `list[str]`.

hopp/simulation/technologies/wind/floris.py
```diff
@@ -239,27 +265,90 @@ def export(self):
...
+            if isinstance(floris_config["farm"]["turbine_type"][0],str):
+                turbine_lib_res = floris_tools.check_libraries_for_turbine_name_floris(floris_config["farm"]["turbine_type"][0], self)
+                if isinstance(turbine_lib_res,str):
+                    raise UserWarning(turbine_lib_res)
+                floris_config["farm"]["turbine_type"][0] = turbine_lib_res
+                
+        else:
+            turbine_lib_res = floris_tools.check_libraries_for_turbine_name_floris(self.config.turbine_name, self)
+            if isinstance(turbine_lib_res,str):
+                turbine_name = check_turbine_name(self.config.turbine_name)
```
RHammond2, line 329: I'm not entirely convinced we should be "fixing" the input in this way. I could be convinced otherwise if there's a good reason we do this with turbine naming conventions but not with other aspects of the model's configuration.

hopp/simulation/technologies/wind/floris.py
```diff
@@ -239,27 +265,90 @@ def export(self):
...
+            # turbine data is included in floris_config
+            if isinstance(floris_config["farm"]["turbine_type"][0],dict):
+                self.turbine_name = floris_config["farm"]["turbine_type"][0]["turbine_type"]
+
+            # load file from internal floris library
+            if isinstance(floris_config["farm"]["turbine_type"][0],str):
+                turbine_lib_res = floris_tools.check_libraries_for_turbine_name_floris(floris_config["farm"]["turbine_type"][0], self)
+                if isinstance(turbine_lib_res,str):
+                    raise UserWarning(turbine_lib_res)
+                floris_config["farm"]["turbine_type"][0] = turbine_lib_res
```
RHammond2, line 324: `floris_config` could be returned after this line, and then the rest of the method could be dedented, and be without an `else` block.

hopp/simulation/technologies/wind/floris.py
```diff
@@ -239,27 +265,90 @@ def export(self):
...
+                turbine_name = check_turbine_name(self.config.turbine_name)
+                turbine_lib_res = get_floris_turbine_specs(turbine_name,self)
+                logger.warning(f"closest matching turbine name to {self.config.turbine_name} is {turbine_name} ... setting turbine model as {turbine_name}")
+            floris_config["farm"]["turbine_type"][0] = turbine_lib_res 
+        return floris_config
+
+    def update_wind_turbine(self,turbine_name):
+        """Update `FlorisModel` (`self.fi`) with with turbine parameters corresponding
+            to `turbine_name`. Used to update turbine parameters after Floris has been initialized.
+            Updates system capacity, rotor diameter, power-curve, and turb_rating.
```
RHammond2, line 338: Can you remove one indent on these two lines? They just need to be in line with the first quote.

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+from hopp.utilities.utilities import write_yaml
+from floris.turbine_library.turbine_previewer import INTERNAL_LIBRARY
+from hopp.utilities import load_yaml
+from hopp.tools.design.wind.turbine_library_interface_tools import get_floris_turbine_specs
+from hopp.tools.design.wind.turbine_library_tools import check_turbine_name, check_turbine_library_for_turbine
+
+def check_output_formatting(orig_dict):
+    """Recursive method to convert arrays to lists and numerical entries to floats.
+        This is primarily used before writing a dictionary to a .yaml file to ensure
+        proper output formatting.
```
RHammond2, line 12: Same comment w.r.t. the description blocck indents at the end of `floris.py`.

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+                    else:
+                        orig_dict[key] = float(val)
+    return orig_dict
+
+def write_floris_layout_to_file(layout_x,layout_y,output_dir,turbine_desc):
+    """Export wind farm layout to floris-friendly .yaml file.
+
+    Args:
+        layout_x (List[float]): x-coordinates of turbines
+        layout_y (List[float]): y-coordinates of turbines
```
RHammond2, line 50: `list[float]`

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+                        orig_dict[k] = new_val
+                    else:
+                        orig_dict[k] = float(val[i])
+            elif isinstance(key,str):
+                if not isinstance(orig_dict[key], (str, bool)):
+                    if isinstance(orig_dict[key], (list, np.ndarray)):
+                        new_val = [float(v) for v in val]
+                        orig_dict[key] = new_val
+                    else:
+                        orig_dict[key] = float(val)
```
RHammond2, line 42: A couple recommended changes below for simplicity  of the control flow. I don't think I quite understand why strings and booleans are added to the end of a list (line 28) versus continuously reset if they aren't lists or arrays (line 35), could explain the rationale here a bit more?

```python
        else:
            if isinstance(key, list):
                for i,k in enumerate(key):
                    if isinstance(orig_dict[k], (str, bool)):
                        orig_dict[k] = orig_dict.get(key, []) + [val[i]]
                    elif isinstance(orig_dict[k], (list, np.ndarray)):
                        orig_dict[k] = np.array(val, dtype=float).tolist()
                    else:
                        orig_dict[k] = float(val[i])
            elif isinstance(key,str):
                if isinstance(orig_dict[key], (str, bool)):
                    continue
                if isinstance(orig_dict[key], (list, np.ndarray)):
                    orig_dict[key] = np.array(val, dtype=float).tolist()
                else:
                    orig_dict[key] = float(val)
```

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+    layout_x = [float(x) for x in layout_x]
+    layout_y = [float(y) for y in layout_y]
+
+    layout = {"layout_x":layout_x,"layout_y":layout_y}
+    n_turbs = len(layout_x)
+    output_fpath = os.path.join(output_dir,f"layout_{turbine_desc}_{n_turbs}turbs.yaml")
+    write_yaml(output_fpath,layout)
+
+def write_turbine_to_floris_file(turbine_dict,output_dir):
+    """Export turbine model to floris-friendly .yaml file.
```
RHammond2, line 64: A minor nitpick, but can you change the reference to YAML files from `.yaml files` to `YAML files`?

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+def check_floris_library_for_turbine(turbine_name):
+    """Check if a turbine exists in the floris internal library.
+
+    Args:
+        turbine_name (str): name of turbine
+
+    Returns:
+        bool: whether turbine exists in floris internal library or not.
+    """
+    floris_library_fpath = INTERNAL_LIBRARY / "{}.yaml".format(turbine_name)
```
RHammond2, line 84: Can you use an  f-string here?

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+    if os.path.isfile(floris_library_fpath):
+        return True
+    else:
+        return False
```
RHammond2, line 88:
```python
return floris_library_fpath.is_file()
```

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+        dict: floris turbine model dictionary
+    """
+    floris_library_fpath = INTERNAL_LIBRARY / "{}.yaml".format(turbine_name)
+    if os.path.isfile(floris_library_fpath):
+        turb_dict = load_yaml(floris_library_fpath)
+        turb_dict.pop("power_thrust_data_file")
+        return turb_dict
+    else:
+        raise FileNotFoundError("Floris library file for turbine {} does not exist.".format(turbine_name))
+
```
RHammond2, line 109:
```python
    if not floris_library_fpath.is_file():
        raise FileNotFoundError("Floris library file for turbine {} does not exist.".format(turbine_name))
    
    turb_dict = load_yaml(floris_library_fpath)
    turb_dict.pop("power_thrust_data_file")
    return turb_dict
```

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+def check_libraries_for_turbine_name_floris(turbine_name,floris_model):
+    is_floris_lib_turbine = check_floris_library_for_turbine(turbine_name)
+    is_turb_lib_turbine = check_turbine_library_for_turbine(turbine_name)
+
+    if is_floris_lib_turbine:
+        turb_dict = load_turbine_from_floris_library(turbine_name)
+        return turb_dict
+    elif is_turb_lib_turbine:
```
RHammond2, line 117: The block above this returns an object, so this `elif` should be an `if`.

hopp/tools/design/wind/floris_helper_tools.py
```diff
@@ -0,0 +1,124 @@
...
+        floris_model.value("turbine_name",turbine_name)
+        turb_dict = get_floris_turbine_specs(turbine_name,floris_model)
+        return turb_dict
+    else:
```
RHammond2, line 121: Same here, so the else can be deleted and the rest of the lines can be dedented once.

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+import matplotlib.pyplot as plt
+import copy
+
+def plot_power_curve(wind_speeds_ms, cp_curve, ct_curve):
+    """Plot Cp and Ct curve per wind speed.
+
+    Args:
+        wind_speeds_ms (List | np.ndarray): list of wind speeds in m/s
+        cp_curve (List | np.ndarray): power curve coefficients (Cp) at each wind speed in `wind_speeds_ms`
+        ct_curve (_type_): thrust curve coefficients (Ct) at each wind speed in `wind_speeds_ms`
```
RHammond2, line 11: `List` -> `list` and `ct_curve` should be of the same argument type as the others.

RHammond2: There are a handful of spots where `List` is used for the docstring typing in this file, can you replace those as well?

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    fig1 = plt.figure()
+    plt.plot(wind_speeds_ms, ct_curve, label="Coeff of Thrust")
+    plt.plot(wind_speeds_ms, cp_curve, label = "Coeff of Power")
+    plt.legend()
+    plt.xlabel("Wind Speed [m/s]")
+    plt.show()
+
+
+def pad_power_curve(wind_speed, curve, v_min = 0.0, v_max = 50.0):
```
RHammond2, line 21: According to the rest of the function, `wind_speed` should be `wind_speed_ms`. I think `wind_speed` would be preferable because m/s is the industry standard, and it's defined in the docstring.

RHammond2: `v_min` and `v_max` are also naming conventions for parts of matplotlib, and aren't being used in that context, could we switch to a more indicative name such as `ws_min`/`ws_max` or `wind_speeed_start`/`wind_speeed_end`?

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+    rotor_area = np.pi*((rotor_diameter/2)**2)
+    if isinstance(wind_speeds_ms, list):
+        wind_speeds_ms = np.array(wind_speeds_ms)
+    if isinstance(power_curve_kw, list):
+        power_curve_kw = np.array(power_curve_kw)
+    
+    # power available in the wind (kW)
+    p_wind = 0.5*air_density*rotor_area*(wind_speeds_ms**3)/1e3
+    cp = power_curve_kw/p_wind
+    cp = np.where(cp < 0, 0, cp)
```
RHammond2, line 83: I would just add a `.tolist()` at the end instead of converting in the return statement.

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+    Raises:
+        UserWarning: if `wind_speeds_ms` and `cp_curve` are different lengths.
+
+    Returns:
+        list: thrust curve coefficients (Ct) at each wind speed in `wind_speeds_ms`
+    """
+    
+    # Check that the wind speed and the coefficient of power are the same length
+    if len(wind_speeds_ms) != len(cp_curve):
+        raise UserWarning("The length of the wind speed and coefficient of power vectors must be the same")
```
RHammond2, line 137: I know the `UserWarning` has been used a lot, but I'm just realizing that it may be being misused. I think it's often being used in place of an actual error quite often, so programs will fail silently where they shouldn't be (such as this specific case). In many cases, I think `ValueError` would be more appropriate unless a non-failing warning is what is intended.

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+    if isinstance(power_curve_kw, list):
+        power_curve_kw = np.array(power_curve_kw)
+    
+    # power available in the wind (kW)
+    p_wind = 0.5*air_density*rotor_area*(wind_speeds_ms**3)/1e3
+    cp = power_curve_kw/p_wind
+    cp = np.where(cp < 0, 0, cp)
+    return list(cp)
+
+def calculate_power_from_cp(wind_speeds_ms, cp_curve, rotor_diameter, rated_power_kW, air_density = 1.225):
```
RHammond2, line 86: In the next few functions, I think `wind_speed_ms` should simply be `wind_speed` as m/s is basically the only form of wind speed that's used in the industry.

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+    ct_curve = list(np.zeros(N_wind))
+    
+    for i in range(N_wind):
+        # calculate induction factor a
+        # solve C_P = 4 * a * (1-a)**a  -> 4 * a**3 - 8 * a**2 + 4 * a - C_P = 0
+        p = np.zeros(4)
+        p[0] = 4
+        p[1] = -8
+        p[2] = 4
+        p[3] = -cp_curve[i]
```
RHammond2, line 149:
```python
roots = np.roots([4, -8, 4, -cp_curve[i]])`
```

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+        print("Coefficient of Thrust (Ct): ")
+        for i in ct_curve:
+            print("-",i)
+        print("Coefficient of Power (Cp): ")
+        for i in cp_curve:
+            print("-",i)
+        print("Wind Speed (v): ")
+        for i in wind_speeds_ms:
+            print("-",i)
+    ct_flat = [ct[0] for ct in ct_curve]
```
RHammond2, line 173: Is this essentially what's happening?
```python
return ct_curve.flatten().tolist()
```

RHammond2, line 173: Is this essentially what's happening?
```python
return ct_curve.flatten().tolist()
```

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,174 @@
...
+    if print_output:
+        print("Coefficient of Thrust (Ct): ")
+        for i in ct_curve:
+            print("-",i)
+        print("Coefficient of Power (Cp): ")
+        for i in cp_curve:
+            print("-",i)
+        print("Wind Speed (v): ")
+        for i in wind_speeds_ms:
+            print("-",i)
```
RHammond2, line 172: This is going to print in a seemingly unhelpful format, could we do something like the following, or similar? Feel free to push back if I'm missing what the intended use is. Just a note that the `:7.4f` is saying left pad to  make the value 7 characters long (or longer if left side of the decimal is more than 3 digits) with 4 decimal accuracy, so feel free to adjust that as needed, if adopting.

```python
print("Wind  Speed (m/s) | Coefficient of Thrust (Ct) | Coefficient of Power (Cp)")
for ws, ct, cp in zip(wind_speeds, ct_curve, cp_curve):
    print(f"{ws:7.4f} | {ct:7.4f} | {cp:7.4f}"
```

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+
+        power_curve_kw = curve_tools.calculate_power_from_cp(wind_speeds,cp_curve,turbine_specs["rotor_diameter"],turbine_specs["rated_power"])
+    elif "power_kw" in turbine_curve_cols and "cp" not in turbine_curve_cols:
+        power_curve_kw = np.array(turbine_specs["power_curve"]["power_kw"].to_list())
+        power_curve_kw = np.where(power_curve_kw<0,0,power_curve_kw)
+        power_curve_kw = np.where(power_curve_kw>turbine_specs["rated_power"],turbine_specs["rated_power"],power_curve_kw)
+        power_curve_kw = list(power_curve_kw)
+        cp_curve = curve_tools.calculate_cp_from_power(wind_speeds,power_curve_kw)
+    else:
+        raise UserWarning(f"turbine {turbine_name} does not have minimum required power curve data (needs either power_kw or cp)")
```
RHammond2, line 44: As it's structured, this is a little hard to follow, so I think 1 of 2 approaches could be taken here. The first is to follow below control flow, which doesn't repeat any pieces of the code, and follows the line of calculating the known, then missing components.

```python
    if "cp" not in turbine_curve_cols and "power_kw" not in turbine_curve_cols:
        raise ValueError(f"turbine {turbine_name} does not have minimum required power curve data (needs either power_kw or cp)")
    
    if "cp" in turbine_curve_cols:
        cp_curve = np.array(turbine_specs["power_curve"]["cp"])
        cp_curve = np.where(cp_curve<0, 0, cp_curve).tolist()
    
    if "power_kw" in turbine_curve_cols:
        power_curve_kw = np.array(turbine_specs["power_curve"]["power_kw"].to_list())
        power_curve_kw = np.where(power_curve_kw<0,0,power_curve_kw)
        power_curve_kw = np.where(power_curve_kw>turbine_specs["rated_power"],turbine_specs["rated_power"],power_curve_kw)
        power_curve_kw = list(power_curve_kw)

    if "power_kw" not in turbine_curve_cols:
        power_curve_kw = curve_tools.calculate_power_from_cp(wind_speeds,cp_curve,turbine_specs["rotor_diameter"],turbine_specs["rated_power"])
    if "cp" not in turbine_curve_cols:
        cp_curve = curve_tools.calculate_cp_from_power(wind_speeds,power_curve_kw)
```

The other is to 1) not mix hte cp_curve and power_curve calculations,  just do one and then the next in a control flow block; and 2) create variables `has_cp` and `has_curve` (or similar) to make the if/elif read more easily.

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+    
+    _, cp_curve = curve_tools.pad_power_curve(wind_speeds,cp_curve)
+    _, ct = curve_tools.pad_power_curve(wind_speeds,ct)
+    wind_speeds, power_curve_kw = curve_tools.pad_power_curve(wind_speeds,power_curve_kw)
+    if model_name == "floris":
+        power_thrust_table = {
+            # "ref_air_density": 1.225,
+            # "ref_tilt": 5.0,
+            # "cosine_loss_exponent_yaw": 1.88,
+            # "cosine_loss_exponent_tilt": 1.88,
```
RHammond2, line 58: If these aren't needed, can we delete them?

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+            "wind_speed":wind_speeds,
+            "power":power_curve_kw,
+            "thrust_coefficient":ct,
+            }
+    elif model_name == "pysam":
```
RHammond2, line 63: Can you update the control flow block to be:

```python
# NOTE: this piece goes at the top
if model_name not in ("floris", "pysam"):
    raise ValueError(<message>)

if model_name == "floris:
    power_thrust_table = ...  # floris stuff
    return power_thrust_table

power_thrust_table = ...  # pysam stuff
return power_thrust_table
```

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+            plant object for either PySAM or FLORIS wind simulation model.
+
+    Returns:
+        float: hub-height to use in meters.
+    """
+    turbine_name = turbine_specs["name"]
+    # if multiple hub height options are available
+    if isinstance(turbine_specs["hub_height"],list):
+        # check for hub height in wind_plant
+        if isinstance(wind_plant,object) and wind_plant is not None:
```
RHammond2, line 101: We should be checking for one of the two options that are specified in the docstring, not `object`. Once this is done, the check for `is not None` can be removed.

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+            
+            # check if hub_height was put in WindConfig
+            if wind_plant.config.hub_height is not None:
+                if any(float(k) == float(wind_plant.config.hub_height) for k in turbine_specs["hub_height"]):
+                    hub_height = wind_plant.config.hub_height
+                    logger.info(f"multiple hub height options available for {turbine_name} turbine. Setting hub height to WindConfig hub_height: {hub_height}")
```
RHammond2, line 107: For longer messages like the ones in this function, could you reformat them to be like the following? The specific line length doesn't matter too much, but a few of these logging messages are well past the 122 character ruler on my monitor, so even with a large monitor in full screen mode it requires a horizontal scroll when viewing two files at once.

```python
msg = (
    f"multiple hub height options available for {turbine_name} turbine."
    f" Setting hub height to WindConfig hub_height: {hub_height}"
)
logger.info(msg)    
```

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+        float: hub-height to use in meters.
+    """
+    turbine_name = turbine_specs["name"]
+    # if multiple hub height options are available
+    if isinstance(turbine_specs["hub_height"],list):
+        # check for hub height in wind_plant
+        if isinstance(wind_plant,object) and wind_plant is not None:
+            
+            # check if hub_height was put in WindConfig
+            if wind_plant.config.hub_height is not None:
```
RHammond2, line 104: This block could be simplified to the following format. Just a note to use the `turbine_hub_height` throughout in place of the dictionary call. I might also just suggest using `name` and `hub_height` because this is a turbine checking function, so the context is fully there already.

```python
# NOTE: this line goes after the turbine_name definition
turbine_hub_height = turbine_specs["hub_height"]

if isinstance(turbine_hub_height,list):
    # check for hub height in wind_plant
    if isinstance(wind_plant,object) and wind_plant is not None:
        if (hub_height := wind_plant.config.hub_height) is not None:
            if any(k == hub_height for k in turbine_hub_height):
                logger.info("...")
                return hub_height
            
        if (hub_height := option2.hub_height) is not None:
            if any(k == hub_height for k in turbine_hub_height):
                logger.info("...")
                return hub_height

    ...  # the other checks
```

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+        power_thrust_table = {
+            "wind_turbine_max_cp": max(cp_curve),
+            "wind_turbine_ct_curve":ct,
+            "wind_turbine_powercurve_windspeeds":wind_speeds,
+            "wind_turbine_powercurve_powerout":power_curve_kw,
+            }
+    return power_thrust_table
+
+
+def check_hub_height(turbine_specs,wind_plant):
```
RHammond2, line 85: Can you update the function definition lines in this file to all have a space after the commas?

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+        hub_height = check_hub_height(turbine_specs,wind_plant)
+        
+        turbine_dict.update({
+            "wind_turbine_rotor_diameter":turbine_specs["rotor_diameter"],
+            "wind_turbine_hub_ht":hub_height,
+            })
+
+    else:
+        raise ValueError(f"turbine {turbine_name} is missing some data, please try another turbines")
+    return turbine_dict
```
RHammond2, line 169: The return line should replace the else statement and the last line of the function should be the error raising.

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,200 @@
...
+            "turbine_type":turbine_name,
+            "hub_height":hub_height,
+            "TSR": turbine_specs["rated_tsr"],
+            "rotor_diameter":turbine_specs["rotor_diameter"],
+            "power_thrust_table": power_thrust_table,
+        }
+        
+    else:
+        raise ValueError(f"turbine {turbine_name} is missing some data, please try another turbines")
+    return turbine_dict
```
RHammond2, line 212: Same comment here about the control flow.

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+
+def load_distributed_turbine_options():
+    """Load list of distributed turbines from turbine-models library.
+
+    Returns:
+        list[str]: list of turbine names in group "distributed"
+    """
+    t_lib = Turbines()
+    distributed_turbines_dict = t_lib.turbines(group = "distributed")
+    distributed_turbines = [v for k,v in distributed_turbines_dict.items()]
```
RHammond2, line 12:
```python
distributed_turbines = list(t_lib.turbines(group="distributed").values())
```

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+
+def load_land_based_turbine_options():
+    """Load list of onshore turbines from turbine-models library.
+
+    Returns:
+        list[str]: list of turbine names in group "onshore"
+    """
+    t_lib = Turbines()
+    lbw_turbines_dict = t_lib.turbines(group = "onshore")
+    lbw_turbines = [v for k,v in lbw_turbines_dict.items()]
```
RHammond2, line 23: Some one-liner as the above function.

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+
+def load_offshore_turbine_options():
+    """Load list of offshore turbines from turbine-models library.
+
+    Returns:
+        list[str]: list of turbine names in group "offshore"
+    """
+    t_lib = Turbines()
+    osw_turbines_dict = t_lib.turbines(group = "offshore")
+    osw_turbines = [v for k,v in osw_turbines_dict.items()]
```
RHammond2, line 34: Same one-liner as the above functions.

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Returns:
+        str: turbine name that most closely matches the input `turbine_name` of the turbines
+             in turbine-models library.
+    """
+    t_lib = Turbines()
+    valid_name = False
+    best_match = ""
+    max_match_ratio = 0.0
+    for turb_group in t_lib.groups:
```
RHammond2, line 51: Should the group be specified as a function argument? There are two possible scenarios I can see where that would be helpful: 1) the user input aligns more with a different group's naming and now they have an offshore turbine for a land-based project (or similar mixed up scenario), and 2) it should be known by the time the model has gotten this far into validation which scenario is being run (unless I'm misunderstanding)

pyproject.toml
```diff
@@ -49,7 +49,8 @@ dependencies = [
     "utm",
     "pyyaml-include",
     "profast",
-    "NREL-rex"
+    "NREL-rex",
+    "turbine_models @ git+https://github.com/NREL/turbine-models.git",
```
RHammond2, line 53: Tagging a git version will render this project unable to be updated on PyPI, so this will either have to get converted to a real package (if there are funds I can help you and Patrick streamline this process, if helpful) or be a recommended git installation for the user.

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+        turbine_name (str): name of turbine in turbine-models library
+
+    Returns:
+        str: turbine name that most closely matches the input `turbine_name` of the turbines
+             in turbine-models library.
+    """
+    t_lib = Turbines()
+    valid_name = False
+    best_match = ""
+    max_match_ratio = 0.0
```
RHammond2, line 50: This seems to be a really low bar to sucess for finding a match.

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+    for turb_group in t_lib.groups:
+        turbines_in_group = t_lib.turbines(group = turb_group)
+        if any(turb.lower()==turbine_name.lower() for turb in turbines_in_group.values()):
+            valid_name = True
+            return turbine_name
+        elif any(turbine_name.lower() in turb.lower() for turb in turbines_in_group.values()):
```
RHammond2, line 56: Because there is a return directly above, this `elif` should be an `if`. I would also change the following since it's duplicated in the next line:

```python
turbine_options = [turb for turb in turbines_in_group.values() if turbine_name.lower() in turb.lower()]
if len(turbine_options) == 1:
    return turbine_options[0]

if len(turbine_options) > 1:
    ... # The code with the else statement (lines 62-66)
```

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+                        best_match = str(turb)
+                        max_match_ratio = max(match_ratio,max_match_ratio)
+
+        else:
+            for turb in turbines_in_group.values():
+                match_ratio = SequenceMatcher(None,turbine_name.lower(), turb.lower()).ratio()
+                if match_ratio>max_match_ratio:
+                    best_match = str(turb)
+                    max_match_ratio = max(match_ratio,max_match_ratio)
+    if valid_name:
```
RHammond2, line 74: This variable is only assigned at the top to False, and just before a return statement, so there is no instance in which this if statement is True.

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+                    if match_ratio>max_match_ratio:
+                        best_match = str(turb)
+                        max_match_ratio = max(match_ratio,max_match_ratio)
+
+        else:
+            for turb in turbines_in_group.values():
+                match_ratio = SequenceMatcher(None,turbine_name.lower(), turb.lower()).ratio()
+                if match_ratio>max_match_ratio:
+                    best_match = str(turb)
+                    max_match_ratio = max(match_ratio,max_match_ratio)
```
RHammond2, line 73: I'm not really convinced that we should be hand holding the user to manage typos other than capitalization and extra spaces because this opens a can of worms of why here and not there. I'm interested to know others' thoughts on this matter because I may be out of touch, but this feels a bit out of scope for this model's goals.

hopp/tools/design/wind/turbine_library_tools.py
```diff
@@ -0,0 +1,94 @@
...
+    if valid_name:
+        return turbine_name
+    else:
+        return best_match
+
+def check_turbine_library_for_turbine(turbine_name:str):
```
RHammond2, line 79: This is functionally the same code as above, just without the `SequenceMatcher` methodology, or substring checking methodology. Personally, this is the version I think is most relevant, though I think we should also be passing which group it comes from.

hopp/tools/resource/wind_tools.py
```diff
@@ -80,8 +80,27 @@ def parse_resource_data(wind_resource):
...
+            wind_dirs = data[:, idx_wd2]
+        
+        else:
+            speeds = data[:, idx_ws].mean(axis=1)
+            wind_dirs = data[:, idx_wd].mean(axis=1)
+    else:
+        # If there's only one hub-height, grab speed and direction data
         speeds = data[:, idx_ws[0]]
         wind_dirs = data[:, idx_wd[0]]
         return speeds, wind_dirs
```
RHammond2, line 106: These should now be in the top-most if-block. The `else` attached to that `if` can now be removed with the code inside that block dedented.

hopp/utilities/utilities.py
```diff
@@ -1,6 +1,6 @@
 import os
 import yaml
-
+import dill
```
RHammond2, line 3: This isn't a current dependency of HOPP, though `multiprocessing-on-dill` is, could you add `dill` since it's clear it should have already been there?

hopp/utilities/utilities.py
```diff
@@ -25,3 +25,29 @@ def load_yaml(filename, loader=Loader) -> dict:
         return filename  # filename already yaml dict
     with open(filename) as fid:
         return yaml.load(fid, loader)
+
+def check_create_folder(filepath):
+    if not os.path.isdir(filepath):
+        os.makedirs(filepath,exist_ok=True)
+        already_exists = False
+    else:
+        already_exists = True
```
RHammond2, line 34: This could be simplified by setting the variable as `True` just before the `if` block.

hopp/utilities/utilities.py
```diff
@@ -25,3 +25,29 @@ def load_yaml(filename, loader=Loader) -> dict:
...
+
+    return already_exists
+
+def write_yaml(filename,data):
+    if not '.yaml' in filename:
+        filename = filename +'.yaml'
+
+    with open(filename, 'w+') as file:
+        yaml.dump(data, file,sort_keys=False,encoding = None,default_flow_style=False)
+    return filename
```
RHammond2, line 44: Is there a reason why a file name is being returned? It doesn't seem to be being modified, so it feels redundant.

hopp/utilities/utilities.py
```diff
@@ -25,3 +25,29 @@ def load_yaml(filename, loader=Loader) -> dict:
...
+
+def dump_data_to_pickle(data,filepath):
+    with open(filepath,"wb") as f:
+        dill.dump(data,f)
```
RHammond2, line 48: I'm not a huge fan of pickling objects, could you just confirm if this is for an intermediary step in the process?

hopp/utilities/utilities.py
```diff
@@ -25,3 +25,29 @@ def load_yaml(filename, loader=Loader) -> dict:
...
+
+def load_dill_pickle(filepath):
```
RHammond2, line 50: lol.

RHammond2: This naming will live rent free in my head whenever I open a pickle jar now.

tests/hopp/test_custom_financial.py
```diff
@@ -8,7 +8,6 @@
 from tests.hopp.utils import create_default_site_info, DEFAULT_FIN_CONFIG
 import copy
 import numpy as np
-
```
RHammond2, line 11: Can you put this deleted line back?

hopp/utilities/utilities.py
elenya-grant: this isn't used or integrated into the workflow at the moment, I just like the option to save objects as pickles for debugging or replicating purposes.

hopp/tools/design/wind/turbine_library_tools.py
elenya-grant: I've agree - I've removed the usage of this function in the code. Instead of using this function, it will just print all the available turbine names and raise a value error.

hopp/tools/design/wind/turbine_library_tools.py
elenya-grant: same comment about this function not being used anymore.

hopp/tools/design/wind/turbine_library_tools.py
elenya-grant: function removed from workflow

hopp/tools/design/wind/turbine_library_tools.py
elenya-grant: function removed from workflow

hopp/tools/design/wind/turbine_library_tools.py
elenya-grant: I agree - function removed from workflow.

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,212 @@
...
+    if has_cp_curve:
+        cp_curve = np.array(turbine_specs["power_curve"]["cp"].to_list())
+        cp_curve = np.where(cp_curve<0,0,cp_curve).tolist()
+
+    if has_power_curve and has_cp_curve:
+        power_curve_kw = np.array(turbine_specs["power_curve"]["power_kw"].to_list())
+        cp_curve = np.array(turbine_specs["power_curve"]["cp"].to_list())
+        power_curve_kw = np.where(power_curve_kw<0,0,power_curve_kw)
+        power_curve_kw = np.where(power_curve_kw>turbine_specs["rated_power"],turbine_specs["rated_power"],power_curve_kw).tolist()
+        cp_curve = np.where(cp_curve<0,0,cp_curve).tolist()
```
RHammond2, line 45: Can you make it so the `power_curve_kw` and `cp_curve` methods are ordered as one, and then other since they aren't reliant on each other? Otherwise, thanks for cleaning this up so this large control flow block so it' easier to follow, the hybrid approach is clearer than my original suggestion too!

johnjasa (commented):
Thanks for this PR, Elenya! I've added a few small comments here and also pushed some minor changes directly to the PR.

Based on our discussion at the code standup, I'll re-review and give full approval once the turbine-models library is installable via pypi. If that's not a priority for this PR, I'm happy to finalize the review before then!

hopp/simulation/technologies/wind/floris.py
```diff
@@ -239,27 +276,83 @@ def export(self):
...
+
+        Raises:
+            UserWarning: if turbine name does not exactly match a turbine in 
+                the turbine library or floris library.
+        """
+        turbine_lib_res = floris_tools.check_libraries_for_turbine_name_floris(turbine_name, self)
+        self.fi.set(turbine_type=[turbine_lib_res])
+        self.value("wind_turbine_rotor_diameter", turbine_lib_res["rotor_diameter"])
+        self.value("wind_turbine_powercurve_powerout", turbine_lib_res["power_thrust_table"]["power"])
+        self.turb_rating = np.round(max(turbine_lib_res["power_thrust_table"]["power"]), decimals = 1)
```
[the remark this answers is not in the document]
elenya-grant: not really - it just prevents issues when checking if the user input a turbine rating and its not equal to the actual turbines rating because of floating point precision. I haven't seen a turbine with a rating precision below 0.1 kW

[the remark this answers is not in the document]
johnjasa: Got it, understood, makes sense. Thanks!

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -100,15 +107,16 @@ class WindConfig(BaseClass):
...
     def __attrs_post_init__(self):
         if self.model_name == 'floris' and self.timestep is None:
             raise ValueError("Timestep (Tuple[int, int]) required for floris")
 
-        if self.layout_mode == 'boundarygrid' and self.layout_params is None:
-            raise ValueError(
-                "Parameters of WindBoundaryGridParameters required for boundarygrid layout mode"
-            )
+        if self.turbine_rating_kw is None and self.turbine_name is None:
+            if self.model_name == "pysam" and self.model_input_file is None:
```
johnjasa, line 118: To be clear, this error should only be raised in the pysam case? Is that correct?

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -191,17 +200,79 @@ def __attrs_post_init__(self):
...
+        """Initialize wind turbine parameters for PySAM simulation.
+
+        Raises:
+            ValueError: if invalid turbine name is provided. Print list of valid turbine names before error is raised. 
+            UserWarning: discrepancy in rotor_diameter value
+            UserWarning: discrepancy in hub-height value
+        """
+
+        if self.config.model_name != "pysam":
+            return
```
johnjasa, line 221: Either this logic or the check above it is unnecessary; suggest to remove one of them. The code can't get to this point unless the model_name is pysam already

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -191,17 +200,79 @@ def __attrs_post_init__(self):
...
+        
         if self.config.rotor_diameter is not None:
             self.rotor_diameter = self.config.rotor_diameter
+        
+        if self.config.turbine_name is not None:
+            valid_name = check_turbine_library_for_turbine(self.config.turbine_name)
+            if not valid_name:
+                print_turbine_name_list()
+                ValueError(f"turbine name {self.config.turbine_name} not found in floris or turbine-models library. Please try an available name.")
```
johnjasa, line 230: nitpick: could you please set this as a `msg` (like you do below) across multiple lines, then raise the error with that msg?

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return power_kW.tolist()
+
+def estimate_thrust_coefficient(wind_speed, cp_curve, plot=False, print_output=False):
+    """Calculate thrust coefficient curve (Ct) from power coefficient curve (Cp).
+
+    Args:
+        wind_speed (list | np.ndarray): list of wind speeds in m/s
+        cp_curve (list | np.ndarray): power curve coefficients (Cp) at each wind speed in ``wind_speed``
+        plot (bool, Pptional): whether to plot Cp and Ct curve. Defaults to False.
```
johnjasa, line 124:
```suggestion
        plot (bool, Optional): whether to plot Cp and Ct curve. Defaults to False.
```

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,161 @@
...
+    """Plot Cp and Ct curve per wind speed.
+
+    Args:
+        wind_speed (list | np.ndarray): list of wind speeds in m/s
+        cp_curve (list | np.ndarray): power curve coefficients (Cp) at each wind speed in ``wind_speed``
+        ct_curve (list | np.ndarray): thrust curve coefficients (Ct) at each wind speed in ``wind_speed``
+    """
+    fig1 = plt.figure()
+    plt.plot(wind_speed, ct_curve, label="Coeff of Thrust")
+    plt.plot(wind_speed, cp_curve, label = "Coeff of Power")
```
johnjasa, line 14:
```suggestion
    plt.plot(wind_speed, cp_curve, label="Coeff of Power")
```

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,161 @@
...
+    Args:
+        wind_speed (list | np.ndarray): list of wind speeds in m/s
+        curve (list | np.ndarray): curve data at each wind speed in ``wind_speed``. Can be 
+            either Cp, power, or Ct.
+        ws_min (float, Optional): wind speed to start curve data at. Defaults to 0.0.
+        ws_max (float, Optional): wind speed to end curve data at. Defaults to 50.0.
+
+    Returns:
+       2-element tuple containing
+
```
johnjasa, line 32:
```suggestion
```

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,242 @@
...
+
+    if has_power_curve and has_cp_curve:
+        power_curve_kw = np.array(turbine_specs["power_curve"]["power_kw"].to_list())
+        power_curve_kw = np.nan_to_num(power_curve_kw)
+        power_curve_kw = np.where(power_curve_kw<0,0,power_curve_kw)
+        power_curve_kw = np.where(power_curve_kw>turbine_specs["rated_power"],turbine_specs["rated_power"],power_curve_kw).tolist()
+        
+        cp_curve = np.array(turbine_specs["power_curve"]["cp"].to_list())
+        cp_curve = np.nan_to_num(cp_curve)
+        cp_curve = np.where(cp_curve<0,0,cp_curve).tolist()
```
johnjasa, line 51: This logic (and the same three lines below) are not needed because they've already been ran in the above code block for `if has_cp_curve` (L38). Delete these blocks?

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,242 @@
...
+        power_curve_kw = curve_tools.calculate_power_from_cp(wind_speeds,cp_curve,turbine_specs["rotor_diameter"],turbine_specs["rated_power"])
+    
+    if has_power_curve and not has_cp_curve:
+        power_curve_kw = np.array(turbine_specs["power_curve"]["power_kw"].to_list())
+        power_curve_kw = np.nan_to_num(power_curve_kw)
+        power_curve_kw = np.where(power_curve_kw<0,0,power_curve_kw)
+        power_curve_kw = np.where(power_curve_kw>turbine_specs["rated_power"],turbine_specs["rated_power"],power_curve_kw).tolist()
```
johnjasa, line 63: These four lines could just exist in a code block for `if has_power_curve` since they're repeated in both blocks for `has_cp_curve` and `not has_cp_curve`; is that correct?

In general, I think the logic in these blocks could be cleaned up or simplified; I'd suggest taking another look at this flow to make sure it's accomplishing what you want in the most clear way

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,242 @@
...
+        "wind_turbine_ct_curve":ct,
+        "wind_turbine_powercurve_windspeeds":wind_speeds,
+        "wind_turbine_powercurve_powerout":power_curve_kw,
+        }
+    return power_thrust_table
+
+
+def check_hub_height(turbine_specs, wind_plant):
+    """Check the hub-height from the turbine-library specs against the other possible hub-height entries. 
+    If multiple hub-height options are available from the turbine_specs, it will choose one based on user-input information.
```
johnjasa, line 96: Could you make this sentence more clear? Just by reading this, I'm not sure what user-inputted info controls this. I'm thinking "it" is "this method" but could you also clarify that?

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,242 @@
...
+
```
johnjasa, line 97: If I'm reading the logic in the function correctly, there's an order for priority for hub heights set by the user; is that correct? If so, could you please note that in the docstring, that it goes WingConfig, site, system model, or median? or whatever the correct order is (I might've misinterpreted)

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,242 @@
...
+        hub_height = check_hub_height(turbine_specs,wind_plant)
+        
+        turbine_dict.update({
+            "wind_turbine_rotor_diameter":turbine_specs["rotor_diameter"],
+            "wind_turbine_hub_ht":hub_height,
+            })
+        return turbine_dict
+
+    raise ValueError(f"turbine {turbine_name} is missing some data, please try another turbines")
```
johnjasa, line 198:
```suggestion
    raise ValueError(f"Turbine {turbine_name} is missing some data, please try another turbine")
```

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,242 @@
...
+        turbine_dict = {
+            "turbine_type":turbine_name,
+            "hub_height":hub_height,
+            "TSR": turbine_specs["rated_tsr"],
+            "rotor_diameter":turbine_specs["rotor_diameter"],
+            "power_thrust_table": power_thrust_table,
+        }
+        return turbine_dict
+        
+    raise ValueError(f"turbine {turbine_name} is missing some data, please try another turbines")
```
johnjasa, line 241:
```suggestion
    raise ValueError(f"Turbine {turbine_name} is missing some data, please try another turbine")
```

hopp/tools/design/wind/turbine_library_interface_tools.py
```diff
@@ -0,0 +1,242 @@
...
+        turbine_specs.setdefault("rated_tsr", 8.0)
+        if turbine_specs["rated_tsr"] is None:
+            turbine_specs["rated_tsr"] = 8.0
+
+        power_thrust_table.update({
+            "ref_air_density": 1.225,
+            "ref_tilt": turbine_specs.setdefault("rotor_tilt_angle", 5.0),
+            "cosine_loss_exponent_yaw": 1.88,
+            "cosine_loss_exponent_tilt": 1.88,
+            })
```
johnjasa, line 231: It seems like this function is setting meaningful values or defaults for these parameters; is that correct? If so, could you please make that clear in the docstring?

Is there a chance this behavior would be unexpected by the user and could it lead to confusion?

johnjasa (changes requested):
Thanks for continuing to work on this PR! I like all the checks and helpful user-facing messages you added as part of this.

I've made some comments and suggestions I'd like to see addressed before approving; nothing too big!

hopp/simulation/technologies/wind/wind_plant.py
elenya-grant: correct - with floris a turbine could be defined in the floris config instead.

hopp/tools/design/wind/turbine_library_interface_tools.py
elenya-grant: I added this to the docstring. I don't think this would unexpected by the user and I'd argue that the user would be more confused if they had to input these values. Many of the turbines in the turbine-library don't have the turbine's rated TSR or rotor tilt angle and most power-curves were made with a reference air density of 1.225 kg/m3. In the future - it would be nice if these defaults could be set within FLORIS or be provided in the turbine-models specifications.

hopp/tools/design/wind/power_curve_tools.py
elenya-grant: removing this white space would make the documentation build for this function look incorrect (although this file isn't included in the docs build yet).

hopp/tools/design/wind/turbine_library_interface_tools.py
johnjasa: Love this cleaned up version, thanks!

johnjasa (approved):
Looks great! Thanks for quickly addressing my suggestions.

hopp/utilities/utilities.py
RHammond2: If this isn't used for the workflow, then I would opt towards not supporting pickling in HOPP itself. Maybe @johnjasa  could chime in as a second opinion here.

pyproject.toml
```diff
@@ -49,7 +49,9 @@ dependencies = [
     "utm",
     "pyyaml-include",
     "profast",
-    "NREL-rex"
+    "NREL-rex",
+    "dill",
+    "turbine-models==0.1.0",
```
RHammond2, line 54: This could probably just be ">=" rather than "==" for flexibility when there are further turbine updates.

hopp/simulation/technologies/wind/floris.py
```diff
@@ -100,60 +97,82 @@ def __attrs_post_init__(self):
...
         self.turb_rating = max(self.wind_turbine_powercurve_powerout)
+        
         if self.config.turbine_rating_kw is not None:
             if self.config.turbine_rating_kw != self.turb_rating:
-                raise UserWarning(
+                msg = (
                     f"Input turbine rating ({self.config.turbine_rating_kw} kW) does not match "
                     f"rating from floris power-curve ({self.turb_rating} kW)"
                 )
+                raise UserWarning(msg)
```
RHammond2, line 140: Should this be a `ValueError`?

hopp/tools/design/wind/power_curve_tools.py
RHammond2: This instance was changed, but it seems there are quite a few them hanging around. Could update each to a more appropriate warning or error to ensure the code fails when it's necessary rather than simply warning a user? I'm adding my recommendation to each. Feel free to resolve them as they're dealt with or push back where you think it's appropriate.

hopp/simulation/technologies/wind/floris.py
```diff
@@ -100,60 +97,82 @@ def __attrs_post_init__(self):
...
+        if self.config.rotor_diameter is not None:
+            if self.config.rotor_diameter != self.wind_turbine_rotor_diameter:
+                msg = (
+                    f"Input rotor diameter ({self.config.rotor_diameter}) does not match "
+                    f"rotor diameter from floris config ({self.wind_turbine_rotor_diameter})"
+                )
+                raise UserWarning(msg)
```
RHammond2, line 147: Should this be a `ValueError`?

hopp/simulation/technologies/wind/floris.py
```diff
@@ -100,60 +97,82 @@ def __attrs_post_init__(self):
...
+        if self.config.hub_height is not None:
+            if self.config.hub_height != hub_height:
+                msg = (
+                    f"Input hub-height ({self.config.hub_height}) does not match "
+                    f"hub-height from floris config ({hub_height})"
+                )
+                raise UserWarning(msg)
```
RHammond2, line 154: Should this be a `ValueError`?

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -191,17 +200,80 @@ def __attrs_post_init__(self):
...
+        if self.config.model_name=="pysam":
+            self.initialize_pysam_wind_turbine()
+        
+        
+    def initialize_pysam_wind_turbine(self):
+        """Initialize wind turbine parameters for PySAM simulation.
+
+        Raises:
+            ValueError: if invalid turbine name is provided. Print list of valid turbine names before error is raised. 
+            UserWarning: discrepancy in rotor_diameter value
```
RHammond2, line 216: This should be a `ValueError` now, and likely the following.

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -191,17 +200,80 @@ def __attrs_post_init__(self):
...
+            self.rotor_diameter = turbine_dict["wind_turbine_rotor_diameter"]
+            self.turb_rating = np.round(max(turbine_dict["wind_turbine_powercurve_powerout"]), decimals = 3)
+
+            if self.config.rotor_diameter is not None:
+                if self.config.rotor_diameter != self._system_model.Turbine.wind_turbine_rotor_diameter:
+                    msg = (
+                        f"input rotor diameter ({self.config.rotor_diameter}) does not match rotor diameter "
+                        f"for turbine ({self._system_model.Turbine.wind_turbine_rotor_diameter})"
+                    )
+                    raise UserWarning(msg)
```
RHammond2, line 245: Should this be a `ValueError`?

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -191,17 +200,80 @@ def __attrs_post_init__(self):
...
+                    f"from for turbine ({self._system_model.Turbine.wind_turbine_hub_ht})"
+                )
 
-        if self.config.model_name=="pysam":
-            if self.config.hub_height is not None:
-                self._system_model.Turbine.wind_turbine_hub_ht = self.config.hub_height
-            if self.config.adjust_air_density_for_elevation and self.site.elev is not None:
-                air_dens_losses = calculate_air_density_losses(self.site.elev)
-                self._system_model.Losses.assign({"turb_specific_loss":air_dens_losses})
+                raise UserWarning(msg)
```
RHammond2, line 254: Should this be a `ValueError`?

hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -235,15 +307,17 @@ def num_turbines(self):
...
+            if n_turbines == len(self._layout.parameters.layout_x):
+                self._layout.set_num_turbines(n_turbines)
+            else:
+                if n_turbines != len(self._system_model.value("wind_farm_xCoordinates")):
+                    n_turbs_layout = len(self._system_model.value("wind_farm_xCoordinates"))
+                    msg = (
+                        f"using custom layout and input number of turbines ({n_turbines}) "
+                        f"does not equal length of layout ({n_turbs_layout})"
+                    )
+                    raise UserWarning(msg)
```
RHammond2, line 320: Should this be a `ValueError`?

hopp/tools/design/wind/power_curve_tools.py
```diff
@@ -0,0 +1,161 @@
...
+        # solve C_P = 4 * a * (1-a)**a  -> 4 * a**3 - 8 * a**2 + 4 * a - C_P = 0
+        roots = np.roots([4, -8, 4, -cp_curve[i]])
+
+        # Take root that is in range of a -> [0, 0.5]
+        a = roots[np.where(np.logical_and(roots>= 0, roots<= 0.5))]
+
+        # Calculate C_T = 4 * a * (1-a)
+        ct_curve[i] = np.round(4 * a * (1-a), 4)
+    
+    ct_flat = [ct[0] for ct in ct_curve]
```
RHammond2, line 151: Could you update this to `ct_curve.flatten().tolist()`? The original comment is disconnected?

hopp/tools/design/wind/turbine_library_tools.py
RHammond2: Re-upping this.

hopp/tools/design/wind/floris_helper_tools.py
RHammond2: Re-upping this.

hopp/tools/design/wind/turbine_library_tools.py
elenya-grant: its weird that this doesnt look updated (I removed this function all together). But - to your point about specifying the turbine group - I have two thoughts. 1) the[ turbine-models library](https://nrel.github.io/turbine-models/Distributed.html) defines distributed wind turbines as turbines with a rated power below 500 kW whereas [NREL's ATB](https://atb.nrel.gov/electricity/2024/distributed_wind) has distributed wind turbines with power ratings up to 1MW. Given this discrepancy in classifying "distributed" wind turbines - I think that it could confuse a user if they specify a turbine as "distributed" based on ATB definition but its in the "onshore" category of the turbine-models library. I do think it'd make sense to distinguish whether the turbine is "offshore" but - I'm not sure what other hopp functionality or considerations exist for offshore wind simulations (we can't access offshore wind resource data at the moment and I'm not sure if Orbit is as tied into the workflow as it is in GreenHEART). I will add `turbine_group` as an optional input to the `check_turbine_library_for_turbine()` function for now.

hopp/tools/design/wind/floris_helper_tools.py
elenya-grant: okay! I updated the workflow and ran a little local test to make sure it doesnt break. TBH - I'm struggling to follow the workflow here but - this method is supposed to make a human-readable yaml file for floris inputs (which has lists of numbers, lists of dictionaries, etc). So - it's supposed replicate the original dictionary but convert any numpy array to a list and convert all any values that aren't strings, booleans or integers into floats. I'd be happy to chat more to go over this and find a nicer solution!

RHammond2: This makes sense to me then. If it's causing a headache due to the nuances of FLORIS configs, then my comment can be disregarded.

hopp/tools/design/wind/turbine_library_tools.py
RHammond2: This explanation works for me, thanks for the thoughtful response!

PR #469, opened 2025-04-01
Bug-fix: WindPlant update for checking inputs

elenya-grant:
# Bug-fix: WindPlant update for checking inputs

Added flexibility for user to model a wind turbine that isn't part of the turbine-models library with PySAM.

## PR Checklist

- [x] Documentation
  - [-] Related `docs/` files are up-to-date, or added when necessary
  - [-] Documentation has been rebuilt successfully
  - [-] Examples have been updated

- `hopp/simulation/technologies/wind/wind_plant.py`: `WindConfig` added inputs
  - `store_floris_config_dict`: boolean input for whether to save floris_config dictionary if running FLORIS.
  -  `override_wind_resource_height`: boolean input for whether to ignore a possible discrepancy between wind resource height and wind turbine hub-height. If set to `True`, this means don't redownload wind resource data for the turbine hub-height if wind resource data is for a different height.
  - `recalculate_pysam_powercurve:`: boolean input for whether to recalculate the power-curve based on rotor diameter and rated power or not. When set to `False` - just scales the power-curve so its maximum value is equal to the input turbine rating (this was the old workflow). When set to `True` - recalculates the turbine power-curve using the `calculate_powercurve()` function from PySAM (see the documentation [here](https://nrel-pysam.readthedocs.io/en/main/modules/Windpower.html#PySAM.Windpower.Windpower.Turbine.calculate_powercurve)). Only used if `model_name = 'pysam'`
- `hopp/simulation/technologies/wind/wind_plant.py`: `WindPlant`
  - `initalize_pysam_turbine_from_turbine_library()`: new function that initializes a wind turbine for PySAM from the turbine-models library, this functionality used to exist in `initialize_pysam_wind_turbine()` but was moved to make code easier to follow.
  - `initialize_pysam_wind_turbine()`: updated workflow if not using a turbine from the turbine models library and user-specified turbine specs like hub-height, rotor diameter and turbine-rating.
- `hopp/simulation/technologies/wind/floris.py`
  - `initialize_from_floris`: updated to check for new input `config.override_wind_resource_height` before downloading wind resource data for the turbine hub-height if there's a discrepancy in wind resource height and wind turbine hub-height.
- `hopp/tools/design/wind/turbine_library_interface_tools.py`
  - `check_hub_height()`: updated so that `wind_plant` input can be None so this function can be used outside of `WindPlant`.

### DIFF
--- a/hopp/simulation/technologies/wind/wind_plant.py
+++ b/hopp/simulation/technologies/wind/wind_plant.py
@@ -76,6 +76,8 @@ class WindConfig(BaseClass):
             Defaults to True.
         override_wind_resource_height (bool): Whether to ignore a possible discrepancy in wind resource height 
             and the turbine hub-height. Defaults to False.
+        override_powercurve_recalculation (bool): If False, recalculates the turbine power-curve for the rotor diameter and turbine rating. 
+            If True, only scales turbine power-curve for turbine rated power. Defaults to False. Only used if ``model_name = 'pysam'``
     """
     # TODO: put `resource_parse_method`, `store_turbine_performance_results`, and `verbose` in "floris_kwargs" dictionary
     num_turbines: int = field(validator=gt_zero)
@@ -120,7 +122,8 @@ class WindConfig(BaseClass):
     store_turbine_performance_results: bool = field(default = False)
     store_floris_config_dict: bool = field(default = True)
     override_wind_resource_height: bool = field(default = False)
-    
+    override_powercurve_recalculation: bool = field(default = False)
+
     def __attrs_post_init__(self):
         if self.model_name == 'floris' and self.timestep is None:
             raise ValueError("Timestep (Tuple[int, int]) required for floris")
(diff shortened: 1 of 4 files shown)


hopp/simulation/technologies/wind/wind_plant.py
```diff
@@ -269,20 +276,25 @@ def initialize_pysam_wind_turbine(self):
...
+                    }
+                    wind_resource = self.site.initialize_wind_resource(data)
+                    self.site.wind_resource = wind_resource
+                    self._system_model.value("wind_resource_data", self.site.wind_resource.data)
 
         if self.config.adjust_air_density_for_elevation and self.site.elev is not None:
             air_dens_losses = calculate_air_density_losses(self.site.elev)
             self._system_model.Losses.assign({"turb_specific_loss":air_dens_losses})
+        
+        if self.config.rotor_diameter is not None and self.config.turbine_rating_kw is not None:
```
jaredthomas68, line 295: I'm uncomfortable with having a power curve modification happening because I did set a rotor diameter and rating but not a turbine name. Is this necessary? I have spent many hours trying to track down hidden changes like this.

jaredthomas68 (changes requested):
I think we need to have some discussions about what changes we are ok with happening without the user intentionally specifying that they should happen.

hopp/simulation/technologies/wind/wind_plant.py
elenya-grant: if you update the rotor diameter without updating the power-curve, the only thing that changes is the layout. If you input a turbine rating and don't recalculate the power-curve, hopp just does scales the power-curve so the max value is equal to the turbine rating input. 

I just updated this workflow and added a logger statement if this is used. I also added an input variable so the turbine power-curve doesnt just "recalculated" - it just gets scaled as it used to anyway.

johnjasa (approved):

PR #480, opened 2025-04-28
Update LDES Example

kbrunik:
# Update LDES Example

Added more narrative around the LDES example highlighting specific modifications for modeling LDES in HOPP.

Also, fixed a type in `site_info.py`.

## PR Checklist

- [-] `RELEASE.md` has been updated to describe the changes made in this PR
- [ ] Documentation
  - [-] Docstrings are up-to-date
  - [ ] Related `docs/` files are up-to-date, or added when necessary
  - [ ] Documentation has been rebuilt successfully
  - [x] Examples have been updated
- [-] Tests pass (If not, and this is expected, please elaborate in the tests section)

NA

NA


kbrunik:
@genevievestarke and @jaredthomas68 I updated the example based on my understanding of your work. Please let me know if there are other key parameters that are missing from the explanation or limitations that would be worth highlighting.

examples/inputs/default_fin_config_ldes.yaml
```diff
@@ -13,18 +13,18 @@ system_costs:
     - 0
   om_capacity:
     - 0
-  om_batt_fixed_cost: 42.888
+  om_batt_fixed_cost: 42.888 # VRDB fixed cost
   om_batt_variable_cost:
     - 0
-  om_batt_capacity_cost: 86.559
+  om_batt_capacity_cost: 86.559 #VRDB capacity cost
```
jaredthomas68, line 19: This is overridden by battery_om_per_kw

examples/11-load-following-long-duration-energy-storage.ipynb
```diff
@@ -9,13 +9,6 @@
     "In this example, we will show how to use and modify the long duration energy storage (LDES) parameters in the hybrid plant simulation. The LDES model is built inside of the battery module, so the parameters are very similar to the battery storage model. In this example we will model a vanadium redox flow battery (VRFB), but we have written the model in a general way so that many types of long-duration energy storage can be modeled reasonably easily."
```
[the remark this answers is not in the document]
kbrunik: <p>Just took a look at the code and it seems that if you set <code>round_trip_efficiency</code> or don't set anything it will set the charge and discharge efficiency to half the round trip efficiency. It does look like you can set <code>charge_efficiency</code> and <code>discharge_efficiency</code>  but it doesn't check against round trip efficiency.</p>

examples/inputs/default_fin_config_ldes.yaml
jaredthomas68: We should probably set it to 0 or the same value as `battery_om_per_kw` - the multiple ways /places of setting financial values is a problem we need to fix. Hopefully that will be remedied in H2I

examples/inputs/default_fin_config_ldes.yaml
```diff
@@ -13,18 +13,18 @@ system_costs:
     - 0
   om_capacity:
     - 0
-  om_batt_fixed_cost: 42.888
+  om_batt_fixed_cost: 42.888 # VRDB fixed cost
```
jaredthomas68, line 16: I see this as a duplicate of `battery_om_per_kw` and should probably be set to zero here. Leaving this value was an error on my part. Again, we need to get away from multiple financial interfaces

jaredthomas68 (changes requested):
Just a few minor comments, half of which are my mistakes from before but I think we should discuss.

examples/inputs/default_fin_config_ldes.yaml
kbrunik: Since most values are the same as the other `default_fin_config.yaml` in the examples folder, would it make sense to use the same default config yaml for this examples? It looks like there are a few formatting changes and additional parameters but it seems like it would be okay to update the `default_fin_config.yaml`

jaredthomas68: I'm fine with that

johnjasa (approved):
Good on my end, thanks for the note on performance!

jaredthomas68 (approved):
Looks good, thanks for the changes.

PR #491, opened 2025-06-04
Bring in function from H2I to copy cost info across both input methods

jaredthomas68:
# Bring overwrite cost function from H2I
This PR primarily brings over the `overwrite_fin_values` function from H2I into HOPP since that function only deals with HOPP and the capability was not already in HOPP.

## PR Checklist

- [x] Documentation
  - [ ] Related `docs/` files are up-to-date, or added when necessary
  - [ ] Documentation has been rebuilt successfully
  - [ ] Examples have been updated

- `hopp/simulation/hopp.py`
  - Import `warnings`
  - import `numpy`
  - `Hopp.from_file()`:  call `overwrite_fin_values` when creating `Hopp` object from a file 
  - `overwrite_fin_values`: function added to properly copy inputs from the hopp config to the respective financial models for wind, PV, and battery
- `hopp/simulation/hopp_interface.py`
  - `HoppInterface.reinitialize()`: call `overwrite_fin_values` before calling `from_dict` because `from_dict` is inherited and I was concerned about overriding it.
- `hopp/simulation/technologies/financial/custom_funcancial_model.py`
  - `CustomFinancialModel.nominal_discount_rate`: clarifying comment stating that the nominal discount rate is calculated using the Fisher equation 
- `pyproject.toml`: update `NREL-PySAM` version to 7.0.0

This is related to H2Integrate PR153 (https://github.com/NREL/H2Integrate/pull/153)

### DIFF
--- a/hopp/simulation/hopp.py
+++ b/hopp/simulation/hopp.py
@@ -90,6 +90,39 @@ class Hopp(BaseClass):
 
 
 def overwrite_fin_values(hopp_config):
+    """
+    Overrides specific financial model values in the HOPP configuration with values from the `cost_info` section.
+
+    This function ensures that the financial model values for technologies (e.g., wind, PV, battery) are updated 
+    with the corresponding values provided in the `cost_info` section of the HOPP configuration. If discrepancies 
+    are found between the values in the financial model and the `cost_info`, the financial model values are 
+    overwritten, and a warning is issued to notify the user.
+
+    Args:
+        hopp_config (dict): The HOPP configuration dictionary containing information about technologies, 
+            financial models, and cost information.
+
+            Expected structure:
+            - `technologies`: Contains technology-specific financial models (e.g., wind, PV, battery).
+            - `config`: Contains the `cost_info` section with updated cost values.
+
+    Returns:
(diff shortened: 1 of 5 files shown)


hopp/simulation/hopp.py
```diff
@@ -83,3 +86,192 @@ def to_file(self, output_file_path: str, filetype: str="YAML") -> None:
                 yaml.dump(self.as_dict(), f, default_flow_style=False)
             else:
                 raise ValueError("Supported export filetype is YAML")
+
+
+
+def overwrite_fin_values(hopp_config):
```
johnjasa, line 92: Does this make more sense living outside of the `hopp.py` file and instead in a utilities-focused file? I understand this is related to the high-level HOPP config _and_ financial values so it could fit in a few different places.

johnjasa (approved):
Thanks for this, Jared!
I left one non-blocking question/suggestion.

hopp/simulation/hopp.py
jaredthomas68: I see your point. I left it here because it is part of the loading/initialization process. I could be talked in to something else but am inclined to leave it here for clarity for now.

PR #500, opened 2025-10-22
Pin timezonefinder, consolidate utilities, rename MHKConfig, removed defunct conda build

### LINKED ISSUE #499: Remove `timezonefinder` as a dependency
# Remove `timezonefinder` as a dependency

I had a colleague with some issues installing H2Integrate and HOPP on Windows (conda) due to `timezonefinder` not building.
Here's the output from his terminal:

```
      error: could not open 'build\lib.win-amd64-cpython-311\timezonefinder\inside_polygon_ext.pyd': Permission denied
      [end of output]
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for timezonefinder
Successfully built h2integrate
Failed to build timezonefinder
error: failed-wheel-build-for-install
× Failed to build installable wheels for some pyproject.toml based projects
╰─> timezonefinder
 ```

## Proposed solution

Remove `timezonefinder` as a dependency in HOPP; it appears to only be used in legacy code and could be an optional dependency

johnjasa:
# Pin timezonefinder, consolidate utilities, rename MHKConfig, removed defunct conda build

The title covers it all!
Resolves #499 by pinning a stable version


dakotaramos:
@johnjasa  I was able to reviewed this, only 1 thought / question. I see you removed tools.utils.py and merged that functionality into utilities.utilities.py. However I don't see export_all(), which was in utils.py, in the new version of utilities.py, I'm assuming because it was not actually used anywhere else in the code?

johnjasa:
> @johnjasa I was able to reviewed this, only 1 thought / question. I see you removed tools.utils.py and merged that functionality into utilities.utilities.py. However I don't see export_all(), which was in utils.py, in the new version of utilities.py, I'm assuming because it was not actually used anywhere else in the code?

Yes, exactly! `export_all()` wasn't used anywhere in the code so I removed it.

